PR #10, opened 2016-10-13
EQDKP Plus Export tweaks

Mirzero:
I've modified the EQ DKP Plus XML Export to include some additional fields that the EQ DKP Plus CMS requires to import successfully.

Some of the data is kind of dummy data... the member join time is always the earliest raid start time, and the leave time is always the latest raid join time + 26000.


evil-morfar:
I suppose the only thing this doesn't account for is if someone have been raiding for more than 8 hours. I know it probably doesn't happen, but the optimal solution would base the leave time around the last awarded item.

Mirzero:
It's kind of dummy data anyway, just to let the export work.
I assumed you weren't even tracking the actual join and leave times, so I have no way to passing those down.

PR #14, opened 2017-09-06
Fixed ticket #285

SafeteeWoW:
Fix autopass isssue when user uses different localization with the loot master. subType should use user localization instead of master looter localization

### DIFF
--- a/core.lua
+++ b/core.lua
@@ -613,6 +613,8 @@ function RCLootCouncil:OnCommReceived(prefix, serializedMsg, distri, sender)
 					for ses, v in ipairs(lootTable) do
 						local iName = GetItemInfo(v.link)
 						if not iName then self:Debug(v.link); cached = false end
+						local subType = select(7, GetItemInfo(v.link))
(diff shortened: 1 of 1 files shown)


SafeteeWoW:
The issue is detailed described in ticket [#285](https://wow.curseforge.com/projects/rclootcouncil/issues/285)

SafeteeWoW:
Found another issue that autopass doesn't work if user changes game client language after RCLootCouncil installed. Have pushed a commit to fix it.

SafeteeWoW:
Extra Suggestion: You might also consider to use user localized "subType" instead of master looter "subType" in VotingFrame

SafeteeWoW:
Use a new way to fix the problem. Just replace "subType" in the lootTable by the user localized value. 
This will also fix the Voting Frame subType display problem(No relic type display), caused by different localization between master looter and the user.

SafeteeWoW:
Not sure if there are any problem similar to this. We should not assume the loot master uses the same localization with the user.

evil-morfar:
Awesome - great find!

PR #19, opened 2017-10-12
Globalize history_table to allow plugin to add data into it

SafeteeWoW:


evil-morfar:
When exactly would you add data to the table to ensure it's there when it's getting sent?

SafeteeWoW:
```
on_click = function(self, data)
    currentAwardingGP = data and data.gp or 0
    if RCLootCouncilML.history_table then
       RCLootCouncilML.history_table["GP"] = gp
    end
    RCLootCouncilML.AwardPopupOnClickYes(self, data)
    currentAwardingGP = 0
    if RCLootCouncilML.history_table then
       RCLootCouncilML.history_table["GP"] = nil
    end
end,
```

SafeteeWoW:
Hold on, gonna push a much better solution.

SafeteeWoW:
A better, more universal solution is to add internal messages RCSendCommandPre and RCSendCommandPost instead.

Then I will do sth like:
```
function RCEPGP:OnMessageReceived(msg, ...)
   if msg == "RCSendCommandPre" then
        local target, command, data  = unpack(...)
        if command == "history" then
            local name, history_table = unpack(data)
            history_table["GP"] = currentAwardingGP
       end
   end
```

evil-morfar:
I do not want messages on `:SendCommand()` - it's intended to send out commands as is, and opening up for changing those commands here instead of where they are made is not good imo. My biggest issue with it though, is the extra overhead this adds. It would execute whatever code anyone has in their listeners twice for each and every command sent out.

I'm not a fan of making the `loot_history` table global either. I think the best way to do this is to add a message to `:TrackAndLogLoot()` before the table is sent, something like:

```lua
local history_table = {}
function RCLootCouncilML:TrackAndLogLoot(name, item, responseID, boss, votes, itemReplaced1, itemReplaced2, reason, isToken, tokenRoll, relicRoll)
   ...
   history_table["relicRoll"] = relicRoll	-- New in v2.5+

   addon:SendMessage("RCHistoryTableSend", history_table, name, item, responseID, boss, votes, itemReplaced1, itemReplaced2, reason, isToken, tokenRoll, relicRoll)

   if db.sendHistory then -- Send it, and let comms handle the logging
	addon:SendCommand("group", "history", name, history_table)
   elseif db.enableHistory then -- Just log it
	addon:SendCommand("player", "history", name, history_table)
   end
   -- Reset table
   history_table = {}
end
```

`history_table` could then remain local and be sent with the message. Placing the `SendMessage()` there would allow others to edit the default data added to it, which might be useful. Placing it above the data inserts would force my data through - I can't decide which I prefer as I see pro's and con's for both approaches; I don't like not knowing if my data is there as that could cause problems for future updates - and on the other hand I should probably trust you/others to only make sensible edits to it.

SafeteeWoW:
I think your solution is good for me.

evil-morfar:
Implemented my solution with c3f696364ac09edcda71ca69e77d221bdec0d156.
I'm currently on vacation so haven't been able to test this in game, but it should work as intended.

evil-morfar:
Forgot the table wipe 558fae4c41ad6afce857151467cb29b34d66a5b7

PR #21, opened 2017-10-18
Fix typo in the AWARD_FOR button of the rightclick menu

SafeteeWoW:
This typo is making AWARD_FOR button unusable, or award to the wrong person, or award to not existing person, because it is accessing the global variable ```name```.


evil-morfar:
Great find, thanks!

PR #22, opened 2017-10-18
Enhancement to how RCLootCouncil send responses.

SafeteeWoW:
When GetItemInfo returns nil, this function will fire error of perform arithmetic on nil value.
Some extra nil check should prevent the errors.


SafeteeWoW:
You should cache all equipped item. This is probably the reason why RCLootCouncil does not show the equipped item sometimes.

evil-morfar:
The only time I've ever had problems with that is when starting a session immediately after logging in. I agree there should be an extra check for this case, but I intentionally left it the way it is so I could be notified if it should ever become a problem. It should never send out an empty diff though.

All this could be fixed by simply requesting the item info whenever the users logs in. Doing it in :OnCommReceived() is a waste, as it will be called many times.

SafeteeWoW:
I did attempt to only cache gears on login, but still get error.
```
function RCLootCouncil:CacheAllEquippedGears()
	local cached = true
	for i = INVSLOT_FIRST_EQUIPPED, INVSLOT_LAST_EQUIPPED do
		local iLink = GetInventoryItemLink("player", i)
		if iLink then
			local iName = GetItemInfo(iLink)
			if not iName then cached = false end
		end
	end
	for i = 1, C_ArtifactUI.GetEquippedArtifactNumRelicSlots() or 0 do
		local iLink = select(4, C_ArtifactUI.GetEquippedArtifactRelicInfo(i))
		if iLink then
			local iName = GetItemInfo(iLink)
			if not iName then cached = false end
		end
	end

	if cached then
		equipped_gears_cached = true
	else
		self:ScheduleTimer("CacheAllEquippedGears", 1) -- Try again after 1s
	end
end
```

However, even if ```equipped_gears_cached``` is true, ```GetInventoryItemLink("player", i)``` could returns an incomplete item string that is not cached (an item string without item name)

SafeteeWoW:
I think the best way is to let CreateResponse returns nil if the current item is not cached, because I find ```GetInventoryItemLink("player", i)``` a little bit inconsistent that it sometimes return incomplete item string that won't be cached.

SafeteeWoW:
New feature that uses ```RCTokenIlvl```
If the item in a session is a token, compare with the current gear using the the min item level of the item the token generates.

Right now, if the session item is a mythic tier20 leg, and I am equipping a 930 leg, voting frame shows the difference as -820, because the item level of the token is 110

After this changes, the diff will be shown correctly as 0.

SafeteeWoW:
More error checking in RCLootCouncilML:GetItemsFromMessage
1. Need to make sure we have received at least two arguments from whisper.
   In the live version, receive a simple whisper "1" will fire an error.
2. I am now retry to execute RCLootCouncilML:GetItemsFromMessage if the item info is not cached.

SafeteeWoW:
Another QoL change to RCLootCouncil:
Current in the live version, the information of currently equipped gear is not sent to the loot master, until the raid member responses.

However, I think it will be much better if we can see the info of current gear immediately. 
Some people are really slow to response, even if they have low gear, and I distribute the item before they have responded.

Thus, I send the response immediately when we received the loot table.
And no info of current gear is sent in lootFrame.lua because it is sent already.

SafeteeWoW:
We dont need to send the current gear if autopassed. Dont need unnecessary info in the voting frame.

SafeteeWoW:
Cache all responses before send any command, to prevent extra commands to be sent in case some item needs to be cached.

SafeteeWoW:
These are lots of changes.
TLDR what I have done:

1. RCLootCouncil:CreateResponse now returns nil if fails to GetItemInfo() on the current gear. 
I find that cache all equipped gear on login is not enough to avoid error.
2. RCLootCouncil:OnCommReceived is delayed by 1s if RCLootCouncil:CreateResponse returns nil.
3. When the session item is a token, the gear difference is calculated against the min ilvl of the item the tokens generates instead of the raw ilvl of the token. (for example, if the token is mythic tier20 token, compare against 930 instead of 110)
4. We don't send the info of current equipped gear if the session item is autopassed, because the loot master don't need this information.
5. If not autopassed, we send the info of current equipped gear immediately when we receive the lootTable, and these information are not sent again when the user click the response buttons.    
I think it's better for the loot master to know the gear of raid members ASAP and there is no reason to fetch that info until the user responses.

SafeteeWoW:
I have tested it by "/rc test", "/rc award" and "/rc add".

SafeteeWoW:
```RCLootCouncil:CreateResponse now returns nil```
Not satisfied with this change. Will change this soon.

**Comment by evil-morfar** [comment]
Loads of great ideas, nice work.

There's a few things though:
* I'm pretty sure `GetInventoryItemLink()` should return either nothing or the complete itemlink - maybe this is a WoW bug?
* I also haven't seen any documentation that `GetInventoryItemLink()` gets cached by the game (although one would assume it will). Only `GetItemInfo()` has been confirmed to do it, so perhaps the link needs to get passed through that as well.
* Even if that's true, do we even need the name of the item? A candidates' equipped item(s) is used for two things: in the voting frame and for the loot history if awarded. The former uses GameTooltip and is requested from the server anyway, and the latter could be gathered before sending out the history.

---
I like what you did with token ilvl - that's how I intended to do it as well.

---
I wasn't aware that could fail, but then again, I haven't tested that for quite a while. Great find though.

---
I had already planned to send out a player's gear along with the loot_ack, where I envisioned appending a table of sessions with the equipped gear for each session. I planned it that way to avoid sending out multiple responses, but I'm not sure it's really required, and your solution would be backwards compatible as well.

I'm adding some comments in the lines of the commit as that'll be easier to explain.

Also, please don't assume what the ML needs. There's no harm in sending the data even on autopasses, and there's certainly no reason to take away data from the user. If nothing else, the info could potentially be useful for the loot history in case of a disenchant award, or if the ML just want to award it to the most deserving in case everyone passes.

**Comment by SafeteeWoW** [comment]
1. I am pretty sure GetInventoryItemLink() has some issue. So I have done extra checks in CreateResponse. Retry to send response after 1s if uses a uncached item.
2. The item level of the token: "110" is never the expected "ilvl" of the token for anyone. So I think this change makes sense.
3. I have made it to send gear info again even if the item is autopassed, as you suggested.

**Comment by SafeteeWoW** [comment]
TLDR what I have done now:

1. RCLootCouncil:CreateResponse now returns nil if fails to GetItemInfo() on the current gear.
New function RCLootCouncil:SendResponse to send response command which automatically delayed by 1s if CreateResponse uses an unused item.
2. More error checkings in RCLootCouncilML:GetItemsFromMessage
3. When the session item is a token, the gear difference is calculated against the min ilvl of the item the tokens generates instead of the raw ilvl of the token. (for example, if the token is mythic tier20 token, compare against 930 instead of 110)
4. we send the info of current equipped gear immediately when we receive the lootTable, and these information are not sent again when the user click the response buttons.

**Comment by SafeteeWoW** [comment]
More Ideas:
1. The default color of relic responses should not be grey.
2. In the response buttons(lootFrame.lua), the item level should be displayed on the left side of the frame. The type of the item( Hands, mail, etc,) should also be added in the buttons. Too many mouseovers on item icons before clicks the buttons.

**Comment by SafeteeWoW** [comment]
Going to bed soon. Can you help me to test what I have commited if you have time? Thanks.

**Comment by evil-morfar** [comment]
Apparently my review didn't post properly when I created it. Not sure how the recent changes affects them, but do have a look.

I don't really care about what color they are - I just left them grey as a reminder to the user that they are not setup by default.
I'd be happy to do a redesign of the lootframe, however when I made it I found the current solution made most sense as it otherwise would be too big. Feel free to make a suggestion though.

**Comment by evil-morfar** [comment]
I'm currently on vacation, so I won't be able to test stuff until monday.

File: core.lua [hunk 1]
```diff
@@ -1007,6 +1028,43 @@ function RCLootCouncil:GetArtifactRelics(link)
...
+function RCLootCouncil:GetRealIlvl(link, rawIlvl)
+	local id = self:GetItemIDFromLink(link)
+	local rawIlvl = rawIlvl or select(4, GetItemInfo(link))
+	local baseIlvl = RCTokenIlvl[id] -- ilvl in normal difficulty
+
+	if not baseIlvl then -- Not token without item level stored
+		return rawIlvl
+	end
+
+	-- Get bonus ids. See epgp/LibItemUtils-1.0.lua
```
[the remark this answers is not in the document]
SafeteeWoW: Resolved by the recent commits.

core.lua
```diff
@@ -669,7 +669,13 @@ function RCLootCouncil:OnCommReceived(prefix, serializedMsg, distri, sender)
 					if GetNumGroupMembers() >= 8 and not IsInInstance() then
 						self:DebugLog("NotInRaid respond to lootTable")
 						for ses, v in ipairs(lootTable) do
-						 	self:SendCommand("group", "response", self:CreateResponse(ses, v.link, v.ilvl, "NOTINRAID", v.equipLoc, nil, v.subType))
+							local session, playerName, responseData = self:CreateResponse(ses, v.link, v.ilvl, "NOTINRAID", v.equipLoc, nil, v.subType)
+							if session then
+						 		self:SendCommand("group", "response", session, playerName, responseData)
+						 	else
+						 		self:Debug("Some items wasn't cached, delaying loot by 1 sec")
+						 		return self:ScheduleTimer("OnCommReceived", 1, prefix, serializedMsg, distri, sender)
```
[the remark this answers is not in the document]
SafeteeWoW: Resolved by recent commits

core.lua
```diff
@@ -1102,27 +1158,49 @@ end
...
+		end
+	end
+
 	return
 		session,
 		self.playerName,
 		{	gear1 = g1,
 			gear2 = g2,
-			ilvl = select(2,GetAverageItemLevel()),
+			ilvl = sendAvgIlvl and select(2,GetAverageItemLevel()) or nil,
```
evil-morfar: Why would you make the average ilvl optional?

SafeteeWoW: If not, will send ilvl once in ML:OnCommReceived and send it again in lootFrame.lua

SafeteeWoW: I don't want to send any data twice.

SafeteeWoW: btw, I think ```history_table = {}``` should be replaced by ```wipe(history_table)```, for performance reason?

evil-morfar:
> TLDR what I have done now:
> 
> 1. RCLootCouncil:CreateResponse now returns nil if fails to GetItemInfo() on the current gear.
> New function RCLootCouncil:SendResponse to send response command which automatically delayed by 1s if CreateResponse uses an unused item.
> 2. More error checkings in RCLootCouncilML:GetItemsFromMessage
> 3. When the session item is a token, the gear difference is calculated against the min ilvl of the item the tokens generates instead of the raw ilvl of the token. (for example, if the token is mythic tier20 token, compare against 930 instead of 110)
> 4. we send the info of current equipped gear immediately when we receive the lootTable, and these information are not sent again when the user click the response buttons.

Also, lets keep this PR to this, and have another issue/PR for your other suggestions :)

SafeteeWoW:
Another new feature I want is the replacement string support in "Announce Consideration"
For example, "&s" for the session number, "&i" for the item.
Then the announcement displayed in the live version is "&s. &i"

core.lua
evil-morfar: The reason ilvl is in `:GetPlayerInfo()` is mainly a hack to make the ilvl available for GroupGear, which had the added benefit of adding candidate's ilvl to the votingframe before a response is received. Ideally, that playerinfo is only sent once when the ML initially gathers info about candidates, and thus will not update if/when a player changes their gear. Imo it still makes sense to send it out. One could argue it only needs to be sent once per session instead of on every response (that might be true for other things as well), but that would be a large change to the comm structure.

Not using wipe is intentional. While it has better "performance" (it reuses the table instead of creating a new), my intend is to make sure the old history_table is no longer used, so whoever listens to the message no longer can make any lasting changes to the table.

SafeteeWoW: The other reason is to that I want to prevent people to change gear before sending out the responses. I want to prevent the bahavior to lower the item level to let the ML to think you have higher upgrades. The current system still does not completely prevent the behavior that the raid member can change gear before the session starts.
Do you think it's good to cache ```GetInventoryItemLink()``` when ```ENCOUNTER_END```?

SafeteeWoW:
And I want to send spec id in the response and add a spec column in the voting Frame.
I know this exists in the ExtraUtilities, but this is a really important information to know, and the inspect data is not as reliable as the messages sent through addon channel.

I think spec information should be in the core of RCLootCouncil instead of EU.

evil-morfar:
I see you've redacted your spec suggestion, but I still have a few thoughts on that.

I almost implemented it in RCLootCouncil before deciding to only have it in ExtraUtilities. Although I only use it as a spec icon in EU, I do think it would be useful to have as a core part of the addon, but dropped it again as I only needed it for that one column.

core.lua
evil-morfar: Ahh, I misunderstood it then. Cool enough.

I don't think it's a good idea to do it on `ENCOUNTER_START` as I fear there might be some InCombatLockDown restrictions - that would be to be tested. Otherwise doing it on `ENCOUNTER_END` should be fine though, unless people have some sort of auto gearswitching macro/addon.

SafeteeWoW:
If you think spec is good to know, I will add that data in

core.lua
SafeteeWoW: ```GetInventoryItemLink``` is not combat restricted.

SafeteeWoW:
I cache the equipped ilvl, gears and relics when encounter starts or login.
Now the user sends the information of the start of encounter instead of the current information.

For example, I start a boss fight with 930 equipped. Immediately after the boss fight, I change my gear to lower the avg ilvl to 915. When the ML starts the session, he still sees me as 930 equipped.

If I would add the spec icon, the spec icon would also show the spec at the start of the last encounter.

SafeteeWoW:
Still decides to add the spec icon

SafeteeWoW:
TLDR what I have done so far:

1. CreateResponse() is replaced by SendResponse. It will retry after 1s delay if uses an uncached item.
2. More error checkings in RCLootCouncilML:GetItemsFromMessage
3. When the session item is a token, the gear difference is calculated against the min ilvl of the item the tokens generates instead of the raw ilvl of the token. (for example, if the token is mythic tier20 token, compare against 930 instead of 110)
4. we send the info of current equipped gears/relics/ilvl immediately when we receive the lootTable, and those information are not sent again when the user click the response buttons.
5. ML receives the information of the status of the user at the start of the last encounter(or user login), instead of the current status. For example, I start a boss fight with 930 equipped. Immediately after the boss fight, I change my gear to lower the avg ilvl to 915. When the ML starts the session, he still sees me as 930 equipped.
6. Spec icon is added to the voting frame.

evil-morfar:
When I talked about the spec icon, I ment add it to the playerData, not adding it as a new column. I'm not sure if people would actually like that (I suspect the average user prefers class icon). 
At the very least it should be a toggleable option that switches between class/spec icon, I don't like having both per default.

I also think it's worth it to break the backwards compatibility and just send all the gear/ilvl/data at once along with the lootAck. These changes have effectively doubled the amount of messages sent by each candidate each session (now: 1 lootAck + n gear + n responses = 2n +1, before: 1 lootAck + n gear/response = n +1). While the amount of data in each message have changed as well (probably less Ace-Comm multiparts) I still think it's more efficient to have less overall messages.

SafeteeWoW:
Ok. I will follow your direction, but I will try to keep backward compatibility.

SafeteeWoW:
Should I put "Not In raid" response after "lootack"?

SafeteeWoW:
In the live version, you are assuming "lookAck" always arrive before "autopass" response.
```
			elseif command == "lootAck" then
				local name = unpack(data)
				for i = 1, #lootTable do
					self:SetCandidateData(i, name, "response", "WAIT")
				end
				self:Update()
```

SafeteeWoW:
I may add the option to let the raiders to choose from only show class icon, only show spec icon, or show both.

SafeteeWoW:
Another idea is to replace ```self:ScheduleTimer``` with ```self:ScheduleTimerSafe``` to provide an upper limit to rerun a function.

SafeteeWoW:
And for all functions that may ScheduleTimer itself, I want to add an extra argument retryTimes to avoid any potential infinite loop. Each time of ScheduleTImer, retryTimes is added by 1. And when retryTimes is over a limit, abort the function.

SafeteeWoW:
I don't know why you transmit equipLoc/subType/ilvl in the lootTable. Isn't link good enough?

SafeteeWoW:
1. Is the ```lootTable``` in votingFrame the same table as the ```lootTable``` in core.lua?
2. Any order when RCLootCoucil:OnCommReceived receives lootTable command and when RCVotingFrame:OnCommReceived receives lootTable command?

SafeteeWoW:
Really dislike equipLoc, subType, etc stored in the lootTable.
Although for the backward compatibility, I may want to keep those information in the lootTable.
But I want to make this version compatible with the future version that eliminates those informations in the lootTable.

See the following code.
```
local itemInfoCache = {}
local itemInfoCategories =
{
	id = true,
	name = true, -- The name of the item in the language of itemLink
	ourName = true, -- The name of the item in our locale.
	link = true,
	quality = true, 
	ilvl = true, -- If it is a token, return the min ilvl of the item it creates.
	rawIlvl = true, -- raw ilvl returns by _G.GetItemInfo
	link = true,
	equipLoc = true,
	texture = true,
	boe = true,
	relicType = true,
	token = true,
}

function RCLootCoucil:GetItemInfo(item, category)
	if not item then return end
	local data = itemInfoCache[item]
	if not category or not itemInfoCategories[category] then
		error("Item info category does not exists"..tostring(category), 2)
	end
	if not data then
		return nil
	end
	return data[category]
end

-- Cache the information of the items.
-- @param item the item link
-- @param autoRetry if true, auto retry is the item hasn't been cached by the Blizzard API.
-- @param retryTimes How many times we have retried. This variable is to avoid infinite loop.
-- @return true if the item is cached or has been cached. nil if hasn't been cached.
function RCLootCouncil:CacheItemInfo(item, autoRetry, retryTimes)
	if not retryTimes or type(retryTimes) ~= "number" then retryTimes = 0 end
	if retryTimes >= 10 then return self:Debug("Stuck in infinite loop", "CacheItemInfo", item, autoRetry, retryTimes) end
	if not item then return end
	if itemInfoCache[item] then return true end

	local ourName, link, rarity, ilvl, iMinLevel, type, subType, iStackCount, equipLoc, texture = GetItemInfo(item)
	local id = link and self:GetItemIDFromLink(link)
	local name = self:GetItemNameFromLink(link)
	if link then
		itemInfoCache[item] = itemInfoCache[item] or {}
		local cache = itemInfoCache[item]
		cache.id = id
		cache.name = self:GetItemNameFromLink(link)
		cache.ourName = ourName -- name in our localization. Although we can get item name from item link, name in the item link may be from other localization.
		cache.link = link
		cache.rawIlvl = ilvl
		cache.ilvl = self:GetRealIlvl(link, ilvl)
		cache.subType = subType
		cache.equipLoc = equipLoc
		cache.texture = texture
		cache.boe = self:IsItemBoE(link)
		cache.relicType = id and IsArtifactRelicItem(id) and select(3, C_ArtifactUI.GetRelicInfoByItemID(id)) 
		cache.token = id and RCTokenTable[id]
		return true
	elseif autoRetry then
		self:ScheduleTimer("CacheItemInfo", 1, item, autoRetry, retryTimes + 1)
	end
end
```

SafeteeWoW:
Send player's gears, relics, equipped ilvl, spec in "lootAck".
Backward compatiblity does NOT break in this commit.

SafeteeWoW:
Add the option to toggle spec icon. Spec icon is not shown by default.
I add the "spec" column in the code, it's just show empty if is toggled off.
"class" column now shows spec icon if the user chooses to hide the class icon

SafeteeWoW:
Add some codes to prevent functions that schedule to run itself to stuck in infinite loop

evil-morfar:
This is getting out of control. For the future please create a seperate pull request for each related feature, as I've lost whatever overview I had on these changes. This PR currently have +429 and −91 changes, which is waaaay too much to review.

evil-morfar:
So in an attempt to follow up:
* Putting `NOT_IN_RAID` after `lootAck` just makes the players gear show up in the voting frame, which adds some extra comm messages, but gives the info to the ML, so I suppose that's fine.
* `lootAck` should always arrive before any response (I've never seen it haven't, even under my most extreme tests) but now that the lootAck comm is way bigger I guess it makes sense to add the check. It should cost much to do.
* I'm not sure I want to add any extra columns, including spec. I'm not even sure I like it being an option in the core rclc, but if so, imo it should just replace the class icon. I certainly don't like it being possible to show both at the same time - that's the sort of stuff EU is for. And showing an empty column is certainly out of the picture.
* While I see your reasoning with adding max tries to all timers, do be careful with those. For example `ML:AddItem()` - I've seen that taking more than 10 tries to get the item, which brings me to my next point:
* The reason for having ilvl, texture, equipLoc, etc. on the lootTable is to avoid doing anything complicated on the clients end. With that data provided I don't need to do any `GetItemInfo()` calls on any of the items in the lootTable. Overall this saves a lot of time and it ensures everyone is working on the exact same data at the same time - the only downside being the item's name will follow the ML's locale. I do not see any reason nor benefit in changing that, except for maybe the item name (which people would have complained about by now if it was an issue).
* Kinda related is your change to where `diff` is calculated. Your change forces everyone to not only cache everyone's gear, but also calculate the diff. By doing it the way I did it would be calculated once, no caching happens, and it would only cost a single integer (well, and it's hashname) in each response. I see no reason nor benefit in changing that.
* It also seems like you're sending ALL the candidates gear (again, the amount of changes makes it hard for me to follow). On the subject of lessening the comm data, why would that even be needed? Only the gear that's in the current session is relevant.
* Why would you want to store all of a candidates gear in the history? (hint: not gonna happen).

Just to clarify when talking about caching: in the live version only the ML needs to cache items (with the exception of relics as mentioned in the comment in `:OnCommReceived()`). Everything is delivered ready for computation. The only exception for this is player gear. We use that for 2 things: the texture and the tooltip. The texture currently requires a call to `GetItemInfo()` thus needs caching, but that could be gathered directly with `GetItemInfoInstant()`. The tooltip have builtin caching, so that doesn't matter.

evil-morfar:
Well, I just ment seperate PR's next time there's massive changes :)

`select(3, C_ArtifactUI.GetRelicInfoByItemID(itemID))` (relic type) is the value that gets stored in the lootTable by the ML, and as mentioned in my note, I don't know if this is the english value (which I can't test - but you might be able to?), which is needed for autopassing. If it's english, then there's no need to cache it. If not, then we need to do the same thing as with `subTypeLookup`.

PR #24, opened 2017-10-21
Enhancement how response is sent (short version)

SafeteeWoW:


SafeteeWoW:
Changelog:
+ User sends ML the info of gears/relics equipped at login or the start of the most recent encounter, instead of those equipped at the time of user response.
  + It is to discourage raid members to equip low ilvl gears to show ML he will get higher upgrade.
+ ML now receives the info of gears/relics from the raid members soon after the start of session, before the raid member responses.
+ If the session item is a token, the ilvl difference is calculated against the min ilvl of the gear it generates, instead of the raw ilvl of the token.
###### Bugfixes
+ Fix an error occurs when the ML starts a session immediately after reload.
+ Fix an error occurs when the ML receives an integer from whisper during a session with whisper feature enabled.

Other changes:
+ Add more items for the test. At least 1 item for each type.

API changes:
###### New
+ ```:UpdatePlayerGears()``` stores the players current gear in ```playersData.gears```
+ ```:UpdatePlayersRelics()``` stores the players current relic in ```playersData.relics```
+ ```:UpdatePlayersRelics()``` calls the above two functions. stores the players current info in ```playersData```
+ ```:SendResponse(target, session, link, ilvl, response, equipLoc, note, subType, isTier, isRelic, sendAvgIlvl)``` sends response to the target.
###### Removed
+ ```:CreateResponse()
###### Changed
```:GetPlayersGear(link, equipLoc, gearsTable)``` adds an arg ```gearsTable```. If specified, use that table to fetch item link instead of from the player's current equipped gears.
```:RCLootCouncil:GetArtifactRelics(link, relicsTable)``` adds an arg ```relicsTable```. Similar to above.

**Comment by SafeteeWoW** [comment]
Changelog:

+ Add an option to show spec icon instead of the class icon in the voting frame.
  + Requires this version to receives the spec information from the user. Still show the class icon if the spec id is not received.

**Comment by SafeteeWoW** [comment]
That's all the features in this PR. no more.

**Comment by SafeteeWoW** [comment]
Some time the lootFrame is not correctly displayed. Not sure if it is caused by my changes:
```
2x ...s\RCLootCouncil\Libs\LibWindow-1.1\LibWindow-1.1-8.lua:128: attempt to perform arithmetic on a nil value
...s\RCLootCouncil\Libs\LibWindow-1.1\LibWindow-1.1-8.lua:128: in function `SavePosition'
RCLootCouncil\core.lua:1915: in function <RCLootCouncil\core.lua:1912>
```
Saved variable:
[RCLootCouncil (2).txt](https://github.com/evil-morfar/RCLootCouncil2/files/1404506/RCLootCouncil.2.txt)

File: core.lua [hunk 1]
```diff
@@ -993,22 +1094,51 @@ function RCLootCouncil:GetPlayersGear(link, equipLoc)
...
 	local id = self:GetItemIDFromLink(link)
 	local g1,g2;
-	if not C_ArtifactUI.GetEquippedArtifactNumRelicSlots() then return end -- Check if we even have an artifact
-	for i = 1, C_ArtifactUI.GetEquippedArtifactNumRelicSlots() do
-		if C_ArtifactUI.CanApplyRelicItemIDToEquippedArtifactSlot(id,i) then -- We can equip it
+	local n = relicsTable and 3 or C_ArtifactUI.GetEquippedArtifactNumRelicSlots() or 0
+	for i = 1, n do
+		local iLink = relicsTable and relicsTable[i] or select(4,C_ArtifactUI.GetEquippedArtifactRelicInfo(i))
+		if iLink and select(3, C_ArtifactUI.GetRelicInfoByItemID(self:GetItemIDFromLink(iLink))) == 
+					 select(3, C_ArtifactUI.GetRelicInfoByItemID(id)) then
```
evil-morfar: Pretty sure `C_ArtifactUI.GetRelicInfoByItemID(id)` requires the incoming relic to be cached - something I'd like to avoid. We should be able to to use the relicType provided by the lootTable for this (still needs locale testing, see comment in #22).

evil-morfar:
This more or less seems fine. I'm home at tuesday, so I probably won't be able to do much until then. I'd like to create a "2.7" branch for this, but I'm not sure if we can change the target of this PR?

Imo it's still worth to consider breaking the backwards compability to avoid having multiple messages sent, though that could easily be changed another time when we actually *need* to break it.

As for the error, I think that's related to my change back to using `UIParent` as the parent for all frames. I think that error was the reason I used `nil` the first time, but I didn't get it when I tested my change. I think it's fixable if we use `f:SetScale()` when the frames are created in `RCLootCouncil:CreateFrame()` before calling `f:RegisterConfig()`, i.e.:

```lua
	local f = CreateFrame("Frame", name, UIParent) -- LibWindow seems to work better with nil parent
	f:Hide()
	f:SetFrameStrata("DIALOG")
	f:SetWidth(450)
	f:SetHeight(height or 325)
        f:SetScale(db.UI[cName].scale or 1.1) -- This should force set the scale to avoid nil arithmatics
	lwin:Embed(f)
	f:RegisterConfig(db.UI[cName])
```
Although it's only a guess.

SafeteeWoW:
+ You can create 2.7 branch and do a pull of the branch Enhancement_how_response_is_sent with a remote of  https://github.com/SafeteeWow/RCLootCouncil2

SafeteeWoW:
Add a table ```RCSubTypeLookup``` that stores ```subType``` for all 11 in-game languages.
Pros:
+ We can now complete avoid caching.
Cons:
+ Painful to update the addon if Blizzard changes translation or we need to add a new entry to track.

```RCSubTypeLookup[GetLocale()]``` is equivalent ```addon.db.global.localizedSubTypes```.
It's your design choice whether you want to keep ```addon.db.global.localizedSubTypes```.

**Comment by SafeteeWoW** [comment]
I doubt if you really want this, so I revert this for now.

**Comment by SafeteeWoW** [comment]
No more caching now, if ML sends english subType.

**Comment by SafeteeWoW** [comment]
Question:
Does ```RCVotingFrame:OnCommReceived``` is always executed after ```RCLootCouncil:OnCommReceived```?
I think it is true because event handler is executed in the order of event registration. Registration happens in ```RCVotingFrame:OnEnable``` and ```RCLootCoucil:OnInitialize``` and the former executes after the latter?

**Comment by SafeteeWoW** [comment]
Fixed RCLootCouncil:GetArtifactRelics and use "relic" entry in the lootTable to avoid caching.

**Comment by SafeteeWoW** [comment]
```RCVotingFrame.SetCellGear``` is using ```GetItemInfo``` to get textures of the gears, so it needs items to be cached. Otherwise, icons are not shown.
To avoid this, the textures of gear1 and gear2 are sent with the response.

**Comment by SafeteeWoW** [comment]
lootTable now shows subType in our localization instead of master looter's localization.

**Comment by SafeteeWoW** [comment]
I am not sure if I should how I handle the item level of the item token.
Instead of calculate the ivl in ```SendResponse```, maybe I should just sends the actual ilvl of the token when ML generates the lootTable (```ML:AddItem```)?

**Comment by evil-morfar** [comment]
From what I'm understanding of AceComm and CallbackHandler, they executed in declaration order, so `RCLootCouncil:OnCommReceived()` will always be called first.

You don't need to send the gear textures. `GetItemInfoInstant()` can deliver those even if the item isn't cached.

It seems like you're trying to add backwards support for when the ML doesn't use v2.7, but the user does. There's no reason to do that - see code comment.

I'm not sure about the token ilvl either, mainly because I still want to save its actual ilvl with the loot history. Doing it in `:AddItem()` would counter that, but it would mean less overall calculations. I think it's best to do it with the response.

File: core.lua [hunk 2]
```diff
@@ -620,6 +629,31 @@ function RCLootCouncil:SendCommand(target, command, ...)
...
+-- @return true if the lootTable is fully localized. false otherwise.
+function RCLootCouncil:LocalizeLootTable(lootTable)
+	-- v2.7 We need to cache if we need to get subType in our locale.
+	-- C_ArtifactUI.GetRelicInfoByItemID() always return english result. So it is not the reason we need to cache.
+	local cached = true
+	for ses, v in ipairs(lootTable) do
+		if v.englishSubType then -- New in v2.7. We don't need to cache if ML sends english subType
+			for subType, englishSubType in pairs(self.db.global.localizedSubTypes) do
+				if englishSubType == v.englishSubType then -- Translate english to our locale
+					v.subType = subType
```
evil-morfar: All this could be replaced with `localizedSubTypesInverse[v.englishSubtype]

core.lua
```diff
@@ -620,6 +629,31 @@ function RCLootCouncil:SendCommand(target, command, ...)
...
+			if not iName then self:Debug(v.link); cached = false end
+			if subType then 
+				v.subType = subType
+				v.englishSubType = self.db.global.localizedSubTypes[subType]
+			end 
+		end
+	end
+	return cached
+end
+
```
evil-morfar: This is adding a new feature to old versions - just show it as before - people need to update to get new features.

evil-morfar: Mainly because it keeps the caching requirement.

core.lua
```diff
@@ -1051,10 +1197,26 @@ local subTypeLookup = {
...
 	["Artifact Relic"]		= 141271, -- Hope of the Forest
+	["Miscellaneous"]       = 151961, -- Legionsteel Flywheel (Trinket)
+	["Junk"]                = 152528, -- Antoran Leggings of the Conqueror (Armor token)
 }
 
 function RCLootCouncil:LocalizeSubTypes()
-	if self.db.global.localizedSubTypes.created == GetLocale() then return end -- We only need to create it once, if game locale is the same as stored locale.
+	if self.db.global.localizedSubTypes.created == GetLocale() then
+		local cached = true
+		localizedSubTypesInverse = {}
```
evil-morfar: `localizedSubTypesInverse` should not be global.

ml_core.lua
```diff
@@ -64,13 +64,14 @@ function RCLootCouncilML:AddItem(item, bagged, slotIndex, index)
...
-		["ilvl"]			= ilvl,
+		["ilvl"]			= addon:GetTokenIlvl(link) or ilvl, -- if the item is a token, ilvl is the min ilvl of the item it creates.
 		["subType"]		= subType,
 		["equipLoc"]	= equipLoc,
 		["texture"]		= texture,
 		["boe"]			= addon:IsItemBoE(link),
 		["relic"]		= itemID and IsArtifactRelicItem(itemID) and select(3, C_ArtifactUI.GetRelicInfoByItemID(itemID)),
 		["token"]		= itemID and RCTokenTable[itemID],
+		["englishSubType"] = addon.db.global.localizedSubTypes[subType],
 	}
```
evil-morfar: Except for backwards compability, is there any reason to add `englishSubType` instead of replacing `subType` with the english?

SafeteeWoW: It's for the backward compatibility.

SafeteeWoW:
What is ```GetItemInfoInstant()```. Where is it declared?

SafeteeWoW:
Why would you want to store the actual ilvl in the loot history?
The actual ilvl of armor token is 110. This is never an item level that makes sense.
This is likely useless information. Furthermore, should the armor tokens in normal/heroic/mythic have the same ilvl in the loot history?

SafeteeWoW:
I have deleted "englishSubtype".
We now never do caching in any case.

SafeteeWoW:
Changelog:
+ User sends ML the info of gears/relics equipped at login or the start of the most recent encounter, instead of those equipped at the time of user response.
  + It is to discourage raid members to equip low ilvl gears to show ML he will get higher upgrade.
+ ML now receives the info of gears/relics from the raid members soon after the start of session, before the raid member responses.
+ RC now determines the ilvl of the armor token as the min ilvl of the item it creates, instead of 110.
+ All raid members should now see the RC Loot Frame opens the same time as the ML.
###### Bugfixes
+ Fix an error when the ML starts a session immediately after reload.
+ Fix an error when the ML receives an integer from whisper during a session with whisper feature enabled.

Other changes:
+ Add more items for the test. At least 1 item for each type.

API changes:
###### New
+ ```:LocalizeLootTable()``` replaces the ```subType``` in the lootTable with the subType in our localization.
+ ```:UpdatePlayerGears()``` stores the players current gear in ```playersData.gears```
+ ```:UpdatePlayersRelics()``` stores the players current relic in ```playersData.relics```
+ ```:UpdatePlayersRelics()``` calls the above two functions. stores the players current info in ```playersData```
+ ```:SendResponse(target, session, link, ilvl, response, equipLoc, note, subType, isTier, isRelic, sendAvgIlvl)``` sends response to the target.
###### Removed
+ ```:CreateResponse()```
###### Changed
```:GetPlayersGear(link, equipLoc, gearsTable)``` adds an arg ```gearsTable```. If specified, use that table to fetch item link instead of from the player's current equipped gears.
```:RCLootCouncil:GetArtifactRelics(link, relicsTable)``` adds an arg ```relicsTable```. Similar to above.

File: core.lua [hunk 6]
```diff
@@ -1204,7 +1185,7 @@ local subTypeLookup = {
 function RCLootCouncil:LocalizeSubTypes()
 	if self.db.global.localizedSubTypes.created == GetLocale() then
 		local cached = true
-		localizedSubTypesInverse = {}
+		local localizedSubTypesInverse = {}
```
evil-morfar: Do we even need this now with the latest changes?

evil-morfar:
[`GetItemInfoInstant()`](https://wow.gamepedia.com/API_GetItemInfoInstant)

I'm not sure why I thought of ilvl in history (that's not getting stored anyway) - I was concerned about altering the value to not match the items' "real" ilvl, but I suppose we only use it for ilvl strings and diff comparisons anyway. Based on that it makes sense to calculate the expected ilvl at the ML.

core.lua
SafeteeWoW: I added two entries "Miscellaneous" and "Junk" into the subTypeLookup( for item type display)
```localizedSubTypesInverse``` is used to make sure those two entries will be lookuped.

SafeteeWoW: If you mean if ```self.db.global.localizedSubTypes``` is needed. yes, we still need it for item type display and autopass, because ```GetItemInfoInstant``` returns localized ```subType```

SafeteeWoW: I don't like this idea. This is forcing caching on the every reload, which is not needed.

evil-morfar: No it doesn't. It would run if `oldVersion < 2.7` = only once. Your way on the other hand loops through `self.db.global.localizedSubTypes` and `subTypeLookup` on every login.

SafeteeWoW: Considering ```self.db.global.localizedSubTypes``` only has 24 entries maximum right now and ```RCLootCouncil:LocalizeSubTypes()``` only runs once on login, looping through it is not an expensive operation needing performance concerns.

SafeteeWoW: 18c7729. Ok. Following your idea.

SafeteeWoW:
Can you help me to resolve this issue?
Got a report on "RCLootCouncil-EPGP" for the capitalization of Russian characters.

wrong:```мяомясо```
correct:```Мяомясо```

I am not sure how to uppercase the first character of Russian.
```name = name:lower():gsub("^%l", string.upper)``` in RCLootCouncil:UnitName is not uppercasing the Russian character correctly. I wonder if you want to fix it, or do not for the backward compatibility in the loot history.

evil-morfar:
Didn't your implementation work for cyrillic characters? My implementation doesn't work for any non ascii characters, and I considered using yours to fix it once and for all, although it probably needs a lot of code to ensure backwards compability.

SafeteeWoW:
Well, figured our I just forget to apply that function.

PR #25, opened 2017-10-22
Fix item type display in the voting Frame, especially for non-English Client

SafeteeWoW:
Changelog:

+ Armor Token is no longer displayed as "Junk" for non-English client.
+ Relic type is now shown for non-English client.
   + Besides, relic is now shown as "xxx Artifact Relic" instead of "xxx" for English client.
+ Trinket is now shown as "Trinket" instead of "Trinket, "Miscellaneous" for non-English client

###### Dev
+ Code that calculates the text of item type is put into a separate function.
+ Fix a usage of ```addon.db.global.localizedSubTypes```
  + new:  ```"Artifact Relic" == self.db.global.localizedSubTypes[subType]```
  + old: ```subType == self.db.global.localizedSubTypes["Artifact Relic"]```
+ "Miscellaneous" is added to the subTypeLookup

### DIFF
--- a/core.lua
+++ b/core.lua
@@ -1051,10 +1051,25 @@ local subTypeLookup = {
 	["Wands"]					= 128096, -- Demonspine Wand
 	["Warglaives"]				= 141604, -- Glaive of the Fallen
 	["Artifact Relic"]		= 141271, -- Hope of the Forest
+	["Miscellaneous"]       = 151961, -- Legionsteel Flywheel (Trinket)
 }
 
 function RCLootCouncil:LocalizeSubTypes()
-	if self.db.global.localizedSubTypes.created == GetLocale() then return end -- We only need to create it once, if game locale is the same as stored locale.
+	if self.db.global.localizedSubTypes.created == GetLocale() then
+		local cached = true
+		localizedSubTypesInverse = {}
+		for sType, name in pairs(self.db.global.localizedSubTypes) do
+			localizedSubTypesInverse[name] = sType
+		end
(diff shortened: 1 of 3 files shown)


SafeteeWoW:
+ Also show the equipment slot of armor tokens

core.lua
```diff
@@ -1917,6 +1933,44 @@ function RCLootCouncil:HideTooltip()
 	GameTooltip:Hide()
 end
 
+-- @return a text of the link explaining its type. For example, "Fel Artifact Relic", "Chest, Mail"
+function RCLootCouncil:GetItemTypeText(link, subType, equipLoc, relicType)
+	local englishSubType = self.db.global.localizedSubTypes[subType]
+
+	local tokenSlot = RCTokenTable[self:GetItemIDFromLink(link)]
+	if tokenSlot then -- It's a token
+		if tokenSlot == "Trinket" then
```
evil-morfar: This `tokenSlot` variable is already delivered with `lootTable.token`.
Isn't it better if a token's equipLoc is added to the lootTable by the ML as well?

evil-morfar:
The artifact relic subType is a sad mistake I really should have discovered a long time ago.
But could you clarify "too many bugs by localization"? Have you noticed other gamebreaking bugs?

SafeteeWoW:
Probably not that many bugs.

core.lua
SafeteeWoW: Gonna change it.

SafeteeWoW:
ML sends the "equipLoc" of armor token now.

PR #26, opened 2017-10-23
Add a feature to request rolls ("/roll") from raid members

SafeteeWoW:
Although RCLootCouncil has a built-in roll feature, I rarely use that. I usually ask raid members to "/roll" for the item.

There are several cons in the RC built-in roll system:
+ ~~~Raid members has no way to validate if the roll result is truly random. Even if the voting frame is shown to every raid member, how they know if the ML manipulates the roll result by changing the source code of the addon?~~~ Not the main reason to  create this feature.
+ More importantly, it feels that ML is controlling the luckiness of the raid members. It is a different feeling to let someone else roll for you, comparing with you rolling for yourself.

Let raid members to enter "/roll" solves this problem. But it is hard for the ML to keep track of roll results in the raid chat. Therefore, I create the feature to request rolls from raid members

+ Add a button "request rolls from raid members" in the voting frame.
+ Once clicked, ML announces the raid "A roll for %s has started. Enter '/roll' without argument to roll for the item. Only the first roll is valid.", and all rolls in the current session are cleared.
+ Once a raid members do the chat command "/roll", the addon parses the message from the event ```CHAT_MSG_SYSTEM``` and show the roll result on the voting frame.

Small changes needed after merged with other PRs


evil-morfar:
I don't think I like this. People are already required to trust the ML to do his job properly, so why shouldn't they trust the rolls? While I understand the concern of the rolls not being especially visible to raiders, I don't  think the extra time it would take and the added complexity is worth it.

SafeteeWoW:
+ I don't always let raiders to "/roll" the items. If no one cares the item, everyone responses "pass" or "transmog", I am letting the addon to do the rolls.
+ I don't like the built-in roll system when the item being rolled is good. I don't want to roll for the raid members when they really care the result of the roll. I want raid members to roll themselves. I don't want to control the luckiness of them.]
+ This feature does not add complexity. You can choose to not use this feature(No complexity added). Or if the ML already prefers to "/roll", this feature actually decreases complexity.

Modules/votingFrame.lua
```diff
@@ -58,6 +60,7 @@ end
 
 function RCVotingFrame:OnEnable()
 	self:RegisterComm("RCLootCouncil")
+	self:RegisterEvent("CHAT_MSG_SYSTEM")
```
evil-morfar: I wonder if it's better to register/unregister the event on `StartManualRoll/EndManualRoll`?

SafeteeWoW:
Fixed. "CHAT_MSG_SYSTEM" is now Registered/unregistered when manual roll starts/ends

SafeteeWoW:
Done testing the feature on all 11 in-game languages.

PR #27, opened 2017-10-23
Enhancement to announcement

SafeteeWoW:
+ Since you are not creating an option (and I agree there is no need to do), it is better to put ```announceItemString``` out of db
+ add ilvl and item type into the announcement.
   + Unfinished. Need small changes after merged with other PRs.


evil-morfar:
Since we're changing it anyway, it might be better to just allow the ML to choose what he wants displayed. 
My original intent was to allow other modules (you) choose what it should print, without changing anything per default as I don't see the need to add more information to it (people with the addon see's all that in their lootframe anyway, so it's mostly an overview and information to potentially anyone without the addon - and they would need to hover over the addon anyway). 

I imagine any modules adding to this would want to make it editable, so based on that it's probably better to just provide the editbox in the core addon.

SafeteeWoW:
Put the desciption of each keyword into separate locale entry.
```
New:
L["announce_awards_desc2"] = "\nChoose which channel(s) you want to announce to along with the text.\n"
L["announce_awards_&i_desc"] = "Use &i for the item awarded."
L["announce_awards_&p_desc"] = "Use &p for the name of the player getting the loot."
L["announce_awards_&r_desc"] = "Use &r for the reason."
L["announce_awards_&n_desc"] = "Use &n for the roll."

Old:
L["announce_awards_desc2"] = "\nChoose which channel(s) you want to announce to along with the text.\nUse &i for the item awarded, &p for the name of the player getting the loot, &r for the reason and &n for the roll."
```

SafeteeWoW:
'&t'(item type) is unfinished. Waiting to merged with #25

PR #28, opened 2017-10-23
Better display in the loot frame

SafeteeWoW:
+ Move ilvl from the topright corner to under the item name
   + I cannot or I don't see ilvl when it is on the topright corner.
+ Show subType
   + Small changes needed after merged with PR #25 
   + With this change, raid members needs to do much less mouseovers to the icon, and less wrong response?
+ lootFrame increases from 75 to 100(33% higher), but I think it's fine?


evil-morfar:
Could you attach an image with an example?

I actually started making a module with various, selectable versions of the loot frame (like minimal, normal, expanded), but I never found a design I actually liked. Maybe yours is good?

Any complains I've heard about regarding the loot frame has always been about it being too huge - I've never actually had any requests to add more info. But if you've made a nice design, I'd happily take it.

SafeteeWoW:
wowscrnshot_102317_101429
wowscrnshot_102317_101435

SafeteeWoW:
Note that the text of item type will be fixed after merged with PR #25

SafeteeWoW:
Frame height increases from 75 to 90, and I decrease the space to the edge.

SafeteeWoW:
wowscrnshot_102317_105547
wowscrnshot_102317_105550

SafeteeWoW:
Tested on other locales. Text space seems okay.

evil-morfar:
Except the note button looks a bit too big, I really like it!
Have you tried reducing the height more? If so, did it affect other locales?

SafeteeWoW:
The title "RCLootCouncil Loot Frame" needs to be moved up with item text and item ilvl in order to reduce the height further.

SafeteeWoW:
wowscrnshot_102317_185044



wowscrnshot_102317_185052

SafeteeWoW:
Frame height is now 85 (13% higher). I moved up the title bar up a bit with the item text.
Further decreasing the size makes the text vertical space too small for some locales.

evil-morfar:
b8890d71cede5b9e465a7c71ad22f6b3dc293b8b
I played around with the spacing as a few things seemed a bit off. First off I wanted to match the note button with the responses, which is now pixel perfect:
image

I also changed the width calculation so the note button has the same spacing from the pass button as the buttons do.

Finally I changed some anchors and most importantly the entry height to 80. As far as my messurements goes there's still the same spacing for the text, but can you confirm it doesn't break on other locales? (I suppose chinese is the only one of concern).
image

PR #30, opened 2017-10-24
Remove unnecessary locale string

SafeteeWoW:
Make use of Global strings in "framexml/ptr/GlobalStrings.lua"
Eliminate unnecessary locale string as much as possible to provide better localization.
Still WIP


SafeteeWoW:
Gonna make a new PR. commits messed up.

PR #32, opened 2017-10-24
Remove unnesessary locale strings

SafeteeWoW:
Using the string constants in "framexml/ptr/GlobalStrings.lua", we can get rid of some locale entries. This will give better translation for some locale, especially if it is not translated on Curseforge yet.

~~~WORK IN PROGRESS.~~~


SafeteeWoW:
I wonder if you like to implement these changes. If so, I will continue.

evil-morfar:
I would very much like to implement this - the fewer translations needed the better.

SafeteeWoW:
It makes some backward compatibility problems to other modules, so some entries are not completed removed.

evil-morfar:
That's fine - we'll just have to remember to update modules before 2.7 goes live.

SafeteeWoW:
Still WIP

SafeteeWoW:
Done

evil-morfar:
I count 54 fewer translations needed - great job!

PR #34, opened 2017-10-25
Adjust title and width of columns

SafeteeWoW:
1. Removed the text "item" at the top of session frame.
2. Replace L["g1"] and L["g2"] by "g1" and "g2". They don't need to be translated. Even if they are translated, the translation cant be displayed within 20 pixel width.
3. Increase the width of several columns in the voting frame to fully display the title of the column for some locales.


evil-morfar:
g1/g2 was intended for if some locale uses a different abbrevation for "gear", such as the russian "п1". It's probably not really relavant, but I think it's better to keep it in.

PR #35, opened 2017-10-25
Fix cloak related issues

SafeteeWoW:
See commit messages. We get cloak autopass issues because ML now sets equipLoc of armor tokens.


evil-morfar:
This doesn't fix backwards compability though.

SafeteeWoW:
ML have to not set equipLoc of armor tokens, in orer to keep compatibility? You want to do that?

SafeteeWoW:
I think we should just break compatibility here.

PR #36, opened 2017-10-25
Show the classes of armor tokens in the voting frame?

SafeteeWoW:
Not sure if this is a good change.
wowscrnshot_102517_152928


evil-morfar:
It's an interesting idea, and worth considering. I just don't think it fits very well with the current layout.

SafeteeWoW:
Will do this in another way. Close for now.

PR #39, opened 2017-10-26
stack duplicate items in the loot frame.

SafeteeWoW:
+ ~~~Add 'Auto Repeat' feature~~~
+ Add 'Stack Items' feature
  + automatically stack duplicate items in the loot session.
  + This often happens when looting the tokens. For example, three heroic vanquisher leg token drops. If I clicked to response either one, auto repeat my responses to the other two.
  + On by default
+ Add 'Print Response' option
  + Print your response in the chat window
  + Off by default
+ Rename 'autopass' option to "Response options"


evil-morfar:
This would still let duplicate items show up in the loot frame. This should only show each item once.

SafeteeWoW:
wowscrnshot_102717_113403

SafeteeWoW:
Now items stack. Duplicate items now only show once in the loot frame.

SafeteeWoW:
wowscrnshot_102717_114420

SafeteeWoW:
wowscrnshot_102717_154747

SafeteeWoW:
wowscrnshot_102717_155737

evil-morfar:
The number on the icon seems kinda redundant (?), but otherwise I like it.

SafeteeWoW:
Probably, but the Blizzard UI shows the stack number on the icon too, so I think it makes sense.

SafeteeWoW:
wowscrnshot_102717_194847

SafeteeWoW:
Removed the item count in the item text.

evil-morfar:
That's a nope. I liked the number in the item text, not the icon. It's too hard to see.

SafeteeWoW:
Reverted it. If you think it's good to show the number in the item text, I think I should also keep the number on the icon because this is how Blizzard UI show stacked items.

evil-morfar:
Was there any particular reason you wanted to print responses? It seems a bit redundant to me.
I also don't think there should be an option to stack the items - a lot of people have been asking for this for a long time, they should just stack all the time.

SafeteeWoW:
+ Print response is for people who likes confirmation and to help people realize he clicked the wrong button.
+ Stack items at all times? sure.

PR #44, opened 2017-10-27
ML sorts lootTable by item type.

SafeteeWoW:
+ ML sorts the lootTable by item type, right before starting a session.


SafeteeWoW:
wowscrnshot_102717_133858

evil-morfar:
Why would you want that?

As ML, I'd prefer to have my list match my input. Other than the visuals, I don't think there's any benefit in doing this.

SafeteeWoW:
> As ML, I'd prefer to have my list match my input.
90% of the time, the input is the loot window from the boss. However, as a ML with RCLootCouncil, I rarely look at the loot window. Even for the BoE I manually added to the session, I don't really care if the session matches the order when I add the BoEs.

> Other than the visuals, I don't think there's any benefit in doing this.
There is benefit. RCLootCouncil automatically switches to the NEXT session after an item has been awarded. This ordering of the lootTable will show items in the same slot consecutively to both ML and other raid members. This helps decision making.
Suppose I am the ML of a guild which distributes all items after full clearing the raid, and there are 30 items in the loot session. With lootTable ordered, ML can consider how to distribute all head pieces as a whole and it's easier for the raid members to response.
Even if there are only 6 items in the session, I think it's super convenient to order the items. Suppose there are two heroic vanquisher tokens drop. After I have awarded one vanquisher token, I really want RCLootCouncil to switch to the session of next vanquisher token, instead of the gear in the next session.

SafeteeWoW:
Gonna make the sorting better.

SafeteeWoW:
```
-- Sorted by:
-- 1. equipment slot: head, neck, ...
-- 2. subType: junk(armor token), plate, mail, ...
-- 3. relicType: Arcane, Life, ..
-- 4. Item level from high to low
-- 5. The value of item bonuses(socket, leech, etc) from high to low
-- 6. Item link
```

SafeteeWoW:
Gonna make it optional(On by default).

SafeteeWoW:
If you think it's needed, I can make an option for this, but I don't think it is necessary.

SafeteeWoW:
wowscrnshot_102717_213852

SafeteeWoW:
+ Added "sortItems" option to Enable/disable this feature.
  + Enabled by default

ml_core.lua
```diff
@@ -151,6 +151,10 @@ function RCLootCouncilML:StartSession()
 	end
 	self.running = true
 
+	if db.sortItems then
+		self:SortLootTable(self.lootTable) -- Sort the lootTable
+	end
+
```
evil-morfar: So the sorting happens AFTER the user starts the session? Aka the order changes from whats seen in the session frame - I really don't like that. The session frame is intended for how the session should look when started (and deleting items, and at some point changing the order manually) - the sorted items needs to be shown the session frame as well.

SafeteeWoW: Changing it.

SafeteeWoW:
Session now is also sorted.

ml_core.lua
SafeteeWoW: Now the sorted items needs to be shown the session frame as well.

SafeteeWoW:
There is a bug when delete an item in the session frame and not all items are loaded.

SafeteeWoW:
Unfinished. Gonna make the code easier.

SafeteeWoW:
Done

### Changelog
+ Now items are sorted by item type and item level by default
   + There is an option to disable this feature.
+ The sorted columns of session frame are reseted whenever it is opened.
+ The item text column of session frame are sorted by item text instead of item link.
####### Bugfixes
+ Fix a bug that when removing item in session frame in test mode, unintended items can be removed if not all items are loaded.
####### Dev
+```ML:AddItem(item, bagged, slotIndex, index)``` is changed to ```ML:AddItem(item, bagged, slotIndex, entry)
  + This changes allows the lootTable to be sorted or has entry removed, when some items are still being loaded.
+ The argument of message "RCMLAddItem" is changed from ```item, session``` to ```item, entry```

SafeteeWoW:
LootTable is sorted ```SessionFrame:Show()``` and ```ML:StartSession()```.
The only problem is that the second one is not needed if not ```db.autoStart```

SafeteeWoW:
Done

### Changelog
+ Now items are sorted by item type and item level by default
   + There is an option to disable this feature.
+ Sorting of session frame is improved.
  + The sorted columns of session frame are reseted whenever it is opened.
  + The item text column of session frame are sorted by item text instead of item link.
  + Delete column is sorted by session number.
  + All columns are sorted ascendingly by default.
###### Bugfixes
+ Fix a bug that when removing item in session frame in test mode, unintended items can be removed if not all items are loaded.
###### Dev
+ ```ML:AddItem(item, bagged, slotIndex, index)``` is changed to ```ML:AddItem(item, bagged, slotIndex, entry)
  + This changes allows the lootTable to be sorted or have its entry removed, when some items are still being loaded.
+ The argument of message "RCMLAddItem" is changed from ```item, session``` to ```item, entry```
----
LootTable is sorted in ```SessionFrame:Show()``` and ```ML:StartSession()```.
The only problem is that the second one is not needed if not ```db.autoStart```

evil-morfar:
I'm gonna postpone this one. Mainly because it's a kinda huge change that I'm not entirely sure how affects everything else, but also because it coincides with the award checking I plan for 2.8.

SafeteeWoW:
I will let the next version of RC-EPGP to support the changes to ```ML:AddItem(item, bagged, slotIndex, index)```

SafeteeWoW:
Need to make changes after merged with #65

Modules/sessionFrame.lua
```diff
@@ -38,9 +38,22 @@ end
 
 function RCSessionFrame:Show(data)
 	self.frame = self:GetFrame()
+	if not self.frame:IsShown() then
```
evil-morfar: There's no reason to do anything with sorting here - it's just visual and isn't used anyway. We don't even have headers on the scrolltable any more, so people actually doing sorting is doing it by mistake. Sorting on the table should really be disabled completely to avoid confusion.

Modules/sessionFrame.lua
```diff
@@ -67,9 +80,9 @@ function RCSessionFrame:ExtractData(data)
 			texture = v.texture or nil,
 			link = v.link,
 			cols = {
-				{ value = "",	DoCellUpdate = self.SetCellDeleteBtn, },
-				{ value = "",	DoCellUpdate = self.SetCellItemIcon},
-				{ value = v.link,	DoCellUpdate = self.SetCellText },
+				{ value = k,	DoCellUpdate = self.SetCellDeleteBtn, },
+				{ value = v.texture or "",	DoCellUpdate = self.SetCellItemIcon},
+				{ value = v.link and addon:GetItemNameFromLink(v.link) or "",	DoCellUpdate = self.SetCellText },
```
evil-morfar: Using itemlink as the text has the added benefit of including the colors of the item - that's the real reason to use it.

evil-morfar: Ah well, forgot this isn't the actual text displayed, derp.

evil-morfar:
Items not being sorted in the session frame could be explained by some items not being cached when `RCSessionFrame:Show()` is called (and it really needs to be, we can't change the order after pressing start). 
But that doesn't explain why `RCLootCouncilML:StartSession()` doesn't do it either. There's no missing values hinting at problems.

image
image

evil-morfar:
It seems to be working as intended now after #85, even though I don't see why. The table should be sorted once when opening the session frame, at which point several items might not be cached. I even confirmed all returns from `GetItemInfo` is nil, but still it manages to get sorted. Beats me 🤔

PR #46, opened 2017-10-27
Differ two %s in the locale string by '%1$s' and '%2$s'. This helps t…

SafeteeWoW:
+ This tells the translator the syntax "%1$s" and "%2$s" exist.


evil-morfar:
Imo this doesn't help, and isn't even worth the relocalization effort.

PR #48, opened 2017-10-31
Localized subtype code cleanup

SafeteeWoW:


evil-morfar:
I don't like this. It forces the `localizedSubTypes` to be recreated on every login, which is basically a waste of cpu (no matter how little impact it has). The `.created` check is a simple solution for that, that even modules can alter if need be.

PR #51, opened 2017-11-03
Better note button

SafeteeWoW:
#### Changelog
+ Add ```L["Click to change your note."]```
+ Note button is desaturated if no note is set.
+ Mouseover any response button also shows note if note is set
  + Allow user to do the final check before sending the response.
+ ```RCLOOTCOUNCIL_LOOTFRAME_NOTE``` dialog is removed.
+ Instead, an editbox is shown when the note button is clicked. The note is saved and editbox is hidden when the user clicks the note button again or presses enter.
   + I think this is much better than a popup because no extra popup that blocks the lootframe and user will always be aware which item he's working on.

wowscrnshot_110217_204750


SafeteeWoW:
Add an maximum limitation of 255 letters, which is the maximum letters allowed by SendChatMessage().
Length beyond this does not make sense.

evil-morfar:
I really like this, nice and clean.

One small issue: each entry retains it's editbox, so if an editbox is open when one clicks a response, it will be open the next time that entry frame is reused. Should be simple to fix, I just don't have more time today.

I could be a bit concerned about the lenght with certain locales, like chinese - does it count each character as 1 or multiple? Otherwise I agree larger notes doesn't make sense.

SafeteeWoW:
+ ```Editbox:SetMaxLetters``` sets the maximum number of letters. 1 Unicode character = 1 letter
  + Currently using this one.
+ ```Editbox:SetMaxBytes``` sets the maximum number of bytes. 1 Unicode character = more than 1 byte.

SafeteeWoW:
> One small issue: each entry retains it's editbox, so if an editbox is open when one clicks a response, it will be open the next time that entry frame is reused. Should be simple to fix, I just don't have more time today.

Fixed.

PR #52, opened 2017-11-03
Allow addon to work without ml

SafeteeWoW:
This resolves Curseforge issues [#134](https://wow.curseforge.com/projects/rclootcouncil/issues/134), [#137](https://wow.curseforge.com/projects/rclootcouncil/issues/137), [#171](https://wow.curseforge.com/projects/rclootcouncil/issues/171)

#### Changelog
+ Before switching loot method to master, check if we can switch to ML method (```addon:CanSetML()```)
+ When the loot method is not master loot, addon sets the "ML" of the addon to be the group leader.
  + People with lower version of RC will be unable to recognize group leader as "ML" and unable to receive command from "ML", but this is not a big compatibility problem. Guild using personal loot are unlikely to use the old version of RC.
+ The group leader can opt in to allow RC to work without ML, which is disabled by default.
#### When the loot method is not master loot and the current player is group leader 
+ Don't start session from loot window
+ Don't adjust loot threshold for auto award.
+ Don't call ```masterlooter``` module unless ```db.workWithoutML``` is true
+ Don't call usage popup unless ```db.workWithoutML``` is true
+ Don't automatically switch loot method to master if ```db.workWithML``` is true
#### Dev
+ New: ```addon:CanSetML()``` returns true if and only if we can set the master looter. (The player is raid leader and in a guild group)
---


SafeteeWoW:
I haven't tested yet, but I think the direction is correct.
Any ideas?

SafeteeWoW:
Haven't tested. will test this tomorrow.

SafeteeWoW:
Updated logic in options.
+ when we allows RC to work without ML, the leader usage should always be enabled unless "Never use RCLootCouncil"
+ When "xxx when leader" is checked, change usage dropdown now does not uncheck it.

You can use "/tinspect RCLootCouncil.db.profile.usage" and then check Dynamic updates, then change usage options in any way to debug it to see if my changes makes senses.

SafeteeWoW:
I think this PR is ready for v2.7

We should give another name to addon.isMasterLooter and addon.masterLooter to avoid confusion. I just don't bother to do that.

SafeteeWoW:
I think this PR is ready for v2.7. Initially, I made this PR complex because I didn't realize I don't need to bother usage if the loot method is not master. Usage is just for loot window, which we don't bother if loot method is not master.

+ Everyone treats group leader as "ML" when the loot method is not master.
+ Don't start session from loot window if loot method is not master.
+ Added a check to see if we can switch loot method to master, so RC doesn't attempt to switch to master loot if the group is not guild group.

We should give another name to addon.isMasterLooter and addon.masterLooter to avoid confusion. I just don't bother to do that.

Don't forget to note this change on the addon frontpage

evil-morfar:
One minor side effect of this is it enables the usage of any ML commands for group leaders when not using master loot, while disabling it until the group leader accepts usage. That's probably fine, as people are used to accept the popup anyway, and probably won't use the commands otherwise.

`addon.masterLooter` is really just a pointer to the "man in charge". The only other name I can think of would be leader, but the ML can also be a nonleader, and it honestly really doesn't matter.

SafeteeWoW:
> One minor side effect of this is it enables the usage of any ML commands for group leaders when not using master loot, while disabling it until the group leader accepts usage. 

Currently, there is an issue in usage. What if user is the ML, but clicks "No" in the usage popup? Then he runs "/rc add". ML will be able to run this command, but because we didn't set ```NewML``` on ml module, there is some issue.

In my opinion, usage should only determine whether or not we start a session from loot, not whether or not we start ml module.

SafeteeWoW:
+ Now ML modules enables if and only if the current player is "the man is charge", regardless of the usage.
+ Usage now only affects whether or not RC starts a session/sessionFrame from loot window. It determines the value of variable ```addon.handleLoot```. It does not affect manual commands such as "/rc award", "/rc add"

core.lua
```diff
@@ -1463,62 +1473,88 @@ end
...
 		return self:ScheduleTimer("NewMLCheck", 2)
 	end
-	if self:UnitIsUnit(old_ml, "player") and not self.isMasterLooter then
-		-- We were ML, but no longer, so disable masterlooter module
+
+	if not self.isMasterLooter then
 		self:GetActiveModule("masterlooter"):Disable()
+	elseif not self:GetActiveModule("masterlooter"):IsEnabled() then
+		self:CallModule("masterlooter")
+		self:GetActiveModule("masterlooter"):NewML(self.masterLooter)
```
evil-morfar: Should probably be moved down, as `self.masterLooter` might still be nil?

core.lua
```diff
@@ -1545,14 +1583,30 @@ function RCLootCouncil:GetML()
 		elseif mlPartyID then		-- Someone in party
 			name = self:UnitName("party"..mlPartyID)
 		end
-		self:Debug("MasterLooter = ", name)
-		-- Check to see if we have recieved mldb within 15 secs, otherwise request it
-		self:ScheduleTimer("Timer", 15, "MLdb_check")
```
evil-morfar: Why remove these?

evil-morfar:
> In my opinion, usage should only determine whether or not we start a session from loot, not whether or not we start ml module.

I agree - this just haven't been an issue before.

PR #53, opened 2017-11-03
Auto add awarded items in bag to the trade window

SafeteeWoW:


SafeteeWoW:
This refers to Curseforge issue [#114](https://wow.curseforge.com/projects/rclootcouncil/issues/114).

+ show dialog to prompt the user when the player trading with is the winner of any awarded items in bag.
+ If click yes, automatically add those items to the trade window.
+ Record any items traded. The items traded to the winner are removed from ```ML..awardedInBags```
  + TODO: print a message to ML or announce to the raid when this happens? Not sure if I should. Add the message if you want.
---
I have tested this feature by playing two characters at the same time. I didn't find a problem even if I 6 items are added together.

SafeteeWoW:
+ Now "/rc add" allows user to enter links without separate them with space.

evil-morfar:
I'm not sure what happened to the space thing between itemlinks - it used to work, but stopped a couple of months back. You really should put that in another pr. Also, is it really that complicated to extract?

SafeteeWoW:
Sure. I have extracted that commit 76f3720 into PR #56 .
It's an easy thing to do with "git cherry-pick"

evil-morfar:
We should really add the `awardedInBags` to the db so that these items are stored if the ML does a /reload. That really completes the whole "store loot and trade later" feature. One more thing to be concerned about is the 2 hour timer on items - I suppose that still applies to these items (I have never tested this in an actual raid).

>print a message to ML or announce to the raid when this happens? Not sure if I should. Add the message if you want.

Shouldn't be necessary.

SafeteeWoW:
Merged into #78

PR #54, opened 2017-11-03
Track all notes in the loot history

SafeteeWoW:
This refers to the Curseforge issue [#306](https://wow.curseforge.com/projects/rclootcouncil/issues/306)
Code is finished and tested.


SafeteeWoW:
I think we'd better record all notes sent by all candidates. I think people are requesting this feature because the guild is using a bidding system, and in bidding system, all notes are very important.
So I want to record all notes.

SafeteeWoW:
+ Instead of only recording the note of winners, record notes of all candidates.
  + Only guilds using bidding system need to record notes, but only recording the note of winners is not very useful for those guild.
  + We should expect only a small portion of raid set note for an item, so we don't need too much space.

SafeteeWoW:
+ Previously, CTRL+Scrollwheel changes the scale of frame, but the size of moreInfo does not change in sync with frame.
+ Now CTRL+Scrollwheel changes the scale of frame and moreInfo

evil-morfar:
I'm not really sure about storing all responses. While it somewhat makes sense to do, we're back to the addon's purpose; which is basically only loot council - with some, over time, added benefits. 

It could probably be displayed nicely in the loothistory in game, but the request was to add it to export, in which it'll be weird.

I updated the tooltip scaling myself.

SafeteeWoW:
I think the main concern is that if we should store all notes, or only store note of winner.

SafeteeWoW:
See the above commit messages for changes.

commas of notes in CSV is replaced by space instead of empty, otherwise, the meaning of sentence could be changed. imagining "11,11" is changed to "1111"

evil-morfar:
That's what I ment by being weird - there's no good way to export all the notes. Only storing the winner is by far the simplest approach.

While I get your point of changing commas to spaces in notes (and maybe responses), it really doesn't make any sense for anything else as those are as close to guaranteed to be proper sentences with space after a comma.

evil-morfar:
That's what I ment by being weird - there's no good way to export all the notes. Only storing the winner is by far the simplest approach.

While I get your point of changing commas to spaces in notes (and maybe responses), it really doesn't make any sense for anything else as those are as close to guaranteed to be proper sentences with space after a comma.

SafeteeWoW:
Used a rebase merge which I shouldn't have done. Should be fine though.

SafeteeWoW:
Commit messes up. Create a new PR

PR #57, opened 2017-11-04
Deprecate RCTokenClasses

SafeteeWoW:
#### Changelog
+ Add function ```RC:GetItemClassesAllowedFlag(item)```. It reads the tooltip to get the classes allowed for the item, as described by the tooltip. This function requires the item to be cached.
  + To save space, it returns a binary flag.
  + Return 0xffffffff if the tooltip does not specify "Classes: xxx"
  + Whether the binary bit `i` is 1 indicates whether the item can be used by the class with id `i`
  + For example, 0x02 (0b10) indicates the item can only be used by class Paladin(class id 2)
  + 0x912 (0b 1001 0001 0010) indicates Paladin, Priest, Warlock, Demon Hunter (classids 2, 5, 9, 12) 
+ Add table ```RC.classDisplayNameToID``` to convert localized class display name to class id
+ Add table ```RC.classTagNameToID``` to convert class name in capital english letters without space to class id (e.g., "WARRIOR', "DEMONHUNTER", etc)
+  ML sends ```classes``` with the return value of ```GetItemClassesAlloedFlag``` in the lootTable. Reroll also sends ```classes```
+ ```RC:AutoPassCheck``` and ```RC:GetItemTypeText``` receives ```classes``` and no longer use ```RCTokenClasses```
+ To preserve backward compatibity, if ```classes``` is not received from lootTable, in ```RC:PrepareLootTable()``` we generate ```classes``` from the data in ```RCTokenClasses```
+ Table ```RCTokenClasses``` is deprecated. This table is now only used in ```RC:PrepareLootTable()``` for backward compatibility purposes.
---
#### Benefit of the changes:
+ No longer need to maintain ```RCTokenClasses```. Less manual work = better.
+ Beside recent set tokens, the addon can now recognize the classes of old raid tokens and the classes of the item created by tokens.
  + For example, if the ML starts the session with a Druid set piece. All other classes will autopass this item. This can be useful if we allow the addon to work without ML.

---
I have tested this PR in all 11 languages and didn't find an issue.
Function ```RC:GetItemClassesAllowedFlag(item)``` is more carefully tested.

---
```bit``` is a built-in bitwise LUA library in WOW. You can get its usage from ```https://github.com/LuaDist/bitlib```


evil-morfar:
Fine idea - but again, I'm not sure LibDeformat is worth it.

SafeteeWoW:
See #73

PR #58, opened 2017-11-04
Use LibDebug to detect leaked globals

SafeteeWoW:
Tired of bugs from leaked globals.
This developer tool will help to detect them

+ LibDebug generates error when we assign global variable, unless we specify ```_G```
+ LibDebug generates error when we read nil global variable, unless we specify```_G```
+ Goto [LibDebug](https://wow.curseforge.com/projects/libdebug) for more information

---
#### Changelog
+ Add LibDebug to embeded libraries.
+ LibDebug is used in every lua file, but flagged by ```--@debug@ --@end-debug@```. 
XML file that loads LibDebug is debug flagged as well, so this change will only be active in the development
This change should **have zero effect to the end user.**
+ Specify ```_G``` to global variables we intend to assign and nil global we intend to read


evil-morfar:
I usually just use https://www.townlong-yak.com/globe/ for checking various stats, but I suppose LibDebug could make sense.

SafeteeWoW:
They do the same stuffs, but LibDebug is more convenient.

PR #59, opened 2017-11-04
Award Improvement and Checking

SafeteeWoW:
~~~+ Dont update loot slots when a session is running or sessionframe is opened.~~~ Will revert it
~~~+ Dont award item if the item is looted.~~~
~~~+ ```ML:UpdateLootSlots()``` is removed~~~
+ WIP

---

~~~I don't see the reason to ```UpdateLootSlots()``` during a session or when session frame is open.LootTable Loot slots should be locked in during a session or session frame~~~
and we should not to award the item if it is already looted.


SafeteeWoW:
+ Add function to parse loot msg

SafeteeWoW:
```
-- The link of same item generated from different player or different spec, are NOT guaranteed to be the same
-- Because item link contains player's level and spec ID.
-- This function compares link with link level and spec ID removed.
-- @return true if two items are the same item
function RCLootCouncil:ItemIsItem(item1, item2)
```
+ Use this function to compare link, unless it's sure that two links are generated at the same time.

SafeteeWoW:
Realized why loot slot needs to be updated, mainly due to Items not master looted in the loot slots.

SafeteeWoW:
WIP. will rewrite this.

evil-morfar:
Just a few comments:
* It really should be enough to check if the item still exists in the loot list after trying to give it out. Could be done either in extension to `:Award()` or by registering an even like `LOOT_SLOT_CLEARED`. Checking chat is slow, combersome and unreliable.
* In general I don't like tests to be included with the addon. I have a seperate addon I use to inject such tests, and alternatively a non packaged file could be used.

SafeteeWoW:
Gonna completely rewrite it. Making a new branch and PR

PR #60, opened 2017-11-05
Use lib deformat for manual roll

SafeteeWoW:
+ Dont know LibDeformat-3.0 until today. Use it to simplify code.


evil-morfar:
Imo it's not worth it to add a whole new Library just to simplify 5 simple lines of code.

PR #62, opened 2017-11-05
Send ml candidates

SafeteeWoW:
+ ML sends master looter candidates (```lootTable.mlCandidates```) with lootTable
+ If a candidate is not on MLcandidate list if the item is master looted, set response as "INELIGIBLE"
+ If the a candidate's response is "INELIGIBLE", its response will always be "ineligible" and cant be changed.
+ backward compatility does not break because we check if ```mlCandidates``` is received.
+ This has been tested by master looting inside a low level dungeon.
---

### Pros
+ "Ineligible" is better information than "not in raid", "offline/RC not installed", "not response in time",

#### Cons 
+ lootTable is bigger, but not too much. For the extreme edge case, 30man in the raid, everyone has 12 letters name and longest realm name, mlCandidates cost 1000 bytes to serialize. ( about the size of lootTable with 4 items, which takes about 1s to transmit) But considering time take in loot session for a big raid, this time does not matter.

### DIFF
--- a/Locale/enUS.lua
+++ b/Locale/enUS.lua
@@ -203,6 +203,7 @@ L["ignore_input_desc"] = "Enter an itemID to add to the ignore list causing RCLo
 L["ignore_input_usage"] = "This function only accepts itemIDs (number)"
 L["ignore_list_desc"] = "Items RCLootCouncil is ignoring. Click on a item to remove it."
 L["ignore_options_desc"] = "Control which items RCLootCouncil should ignore. If you add an item that isn't cached, you need switch to another tab and back before you'll see it in the list."
+L["Ineligible"] = true
 L["Item received and added from 'player'"] = "Item received and added from %s."
 L["Item was awarded to"] = true
 L["Item"] = true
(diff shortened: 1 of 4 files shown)


SafeteeWoW:
After this change, if lootTable is traversed by ```pairs``` instead of ```ipairs```, error will happen, but I don't think ```pairs``` is used in any earlier version of RC for lootTable.

evil-morfar:
I agree with the concept of this, but definitely not with the current solution.

First of all, "Ineligible" should be an addition to the current responses for when the candidate actually is ineligible for receiving loot - all the current responses gives more info as is.
As such, "Ineligible" response must be changeable.

The loottable would not only be bigger, it would be *unnecessary* bigger. A simple improvement to this would be to inverse it, and only send out ineligible names, as they per default will be the minority.

Also, the ineligible status is fluid, and can change everytime the loot list is opened. Imagine someone doing a quick HS after a kill - they would temporarily be ineligible while outside the raid (coincidently an example of where "Not in Raid" is waaay better info), but become eligible again once back.

SafeteeWoW:
Dont know ML candidate list is fluid. Then this PR does not work.

PR #70, opened 2017-11-09
Remove LUA export

SafeteeWoW:
+ Syntax errors in LUA export is fixed.
+ Export is ordered by keys


evil-morfar:
What syntax errors are fixed? I don't remember getting any.

While I agree there could be some cleanup (like not exporting empty names tables) I just never got around to doing it as I think the lua export have become quite redundant, and really was intended as "the first way to export to other people".

I myself do prefer to have a tabbed export, regardless of the ingame representation of it.

Also, these changes makes the export at least 3 times slower (not to mention adding some weird looking code) just to have the names sorted alphabetically - which is broken anyway next time a new candidate is added - all in all not worth it imo.

SafeteeWoW:
+ Current LUA export:
  + No double quote around string
  + itemLink is used as table key. I don't understand that.
+ I agree that LUA export is redundant. Maybe just drop it.

+ Export function that generates export strings shouldn't have any performance concerns.
  + Even if the function runs 100ms, it is okay. No one run export frequently.
  + Yes, huge exports causes lag, but it is caused by the Multiline editbox when it is assigned to a large string.

SafeteeWoW:
+ lua export is removed
+ Default export is set to TSV

evil-morfar:
Making tsv the default is probably a good choice as it's really the most useful one.

I wouldn't go as far as removing the lua export completely, as it's only ~40 lines, and some people might be using it for something.

evil-morfar:
Clearly it was more broken than I thought, so people probably haven't been using it, as it's almost been 6 months since I broke it - removing it.

PR #71, opened 2017-11-10
Huge import/export no longer freezes the game

SafeteeWoW:
### What causes the lag?

+ Editbox widget in multiline mode cant handle huge text properly.
  + See http://www.wowinterface.com/forums/showthread.php?p=315500
+ LootHistory:ExportTSV(), LootHistory:ExportCSV(), etc costs little time, and do not contribute to lag.
  + We don't need to concern about the performance of above functions, really.

---
#### Changelog
+ TLDR: No lag in Import/Export. No matter how large it is.
+ **Export:**
  + When the export is huge, we display the export in a single line editbox to avoid freezing the game
    + Even if only one line is shown, the newline character ```\n``` can still be copied and pasted correctly.
    + This completely removes lag.
  + User can no longer modify text in export editbox.
    + Done by resetting the content of editbox on ```OnTextChanged``` event
  + WeakAuras2 export can cause lag, so I don't use WeakAuras2 code for export
+ **Import:**
  + Import still uses multiline editbox, but only shows first 2500 bytes to avoid freezing the game
    + Credit to WeakAuras 2: See https://github.com/WeakAuras/WeakAuras2/blob/master/WeakAurasOptions/OptionsFrames/ImportExport.lua
    + The editbox is SetMaxBytes to 2500 characters and we use ```OnChar``` event to fetch the characters.
    + Autofocus editbox when import frame opens
+ The width and height of import/export frame is changed.
   + This matches the width/height radio of history frame.
   + The other reason of this change is that I added some description of the editboxes. I need to increase the width to fully show those description


evil-morfar:
I'm aware the editbox causes the lag - my other comment was about I didn't like adding a lot of code for little gain. One thing worth mentioning is the way lua handles strings: creating/changing a lot of them is generally a bad idea, and should be avoided when possible - hence the usage of table.concat.

While I don't really like the aesthetics of a single line, I do think it's very much worth including if it can remove the lag. Maybe it's worth only using the single line version for consistency?

SafeteeWoW:
I like to have both versions.
I like to have multiple line version to enable preview, whenever possible. I don't want user to lose feature that already works fine.
I also want to do large export, so I create single line version.

PR #74, opened 2017-11-11
Ignore options fix

SafeteeWoW:
+ Now we use a sparse table instead of an indexed table to store ignored items.
  + The default state of indexed table cant be handled properly. Imagine in the new version we add an ignored item to the default ignored list, which contained 9 items in the old version. So that item located at ```db.ignore[10]```. The user has an custom ignored item in the old version, which is also at ``db.ignore[10]```. Our update to the default ignored items won't be applied.
+ Ignore option now also accepts itemName and itemLink as input.
  + GetItemInfoInstant does the validation and transfrom the input into itemid
+ Dropdown also shows item id.
+ Added a default ignored item:
  + (id: 155831)  [Pantheon's Blessing]: Use: Increase the power of a Path of the Titans trinket by 5 item levels, up to item level 1000.


evil-morfar:
Didn't think about that - great find.

evil-morfar:
Come to think of it, I'm a bit concerned about people having set new items in here - I think that will cause errors when being combined with the new table structure, so it might require a reset.

PR #75, opened 2017-11-11
Ensure council is up-to-date

SafeteeWoW:
+ ML should send council list when council in group updates.
+ Check if council is received in lootTable

---
I have considered to do sth similar to CandidateCheck in RCVotingFrame, not sure how to do it though.

### DIFF
--- a/core.lua
+++ b/core.lua
@@ -1265,7 +1265,7 @@ function RCLootCouncil:Timer(type, ...)
 				self:SendCommand(self.masterLooter, "MLdb_request")
 			end
 			-- and if we haven't received a council, request it
-			if #self.council == 0 then
+			if #self.council == 0 or not tContains(self.council, self.masterLooter) then
 				self:SendCommand(self.masterLooter, "council_request")
 			end
 		end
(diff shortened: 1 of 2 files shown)


core.lua
```diff
@@ -694,6 +694,19 @@ function RCLootCouncil:OnCommReceived(prefix, serializedMsg, distri, sender)
...
+							mlInCouncil = true
+							break
+						end
+					end
+					if not mlInCouncil then 
+						self:Debug("Received loot table without ml in the council :(", sender)
+						self:SendCommand(self.masterLooter, "council_request")
+						return self:ScheduleTimer("OnCommReceived", 5, prefix, serializedMsg, distri, sender)
+					end
+
```
evil-morfar: If the table exists, then the ML will be in it, so it's a quite redundant check. Could also be shortened to:
```lua
if not tContains(self.council, self.masterLooter) then 
   self:Debug("Received loot table without ml in the council :(", sender)
   self:SendCommand(self.masterLooter, "council_request")
   return self:ScheduleTimer("OnCommReceived", 5, prefix, serializedMsg, distri, sender)
end
```

PR #76, opened 2017-11-11
Security enchancement: only send response and vote to council

SafeteeWoW:
+ Non-council members can easily see the voting frame, by adding one or two lines of code in RC
+ This code makes RC more secure. Send actual response and vote only to council members.

---
#### Potential problems'
+ We must make sure everyone in the group receives up-to-date council list, otherwise we have problems. I have made another PR to do it.
+ We send one message for every council we have, so it has performance problem when there are too many councilmen.
   + However, commands we send to council are all less than 100 bytes, so it should be fine.


evil-morfar:
Alot of stuff can be "hacked" by modifying the code, and it doesn't need to be that secure, after all, the only "real" thing that actually requires this level of secrecy is who's voting for whom, which you can get anyway by sending a ´reconnect` request to the ML (that could of cause be changed). 
You'd also need to add extra code in case the ML adds a councilmember mid session (which actually happens quite often when rotating councilmembers as my guild do).

I also rely on it (/rc nnp) to gather data in raids when I'm not on the council/ML.

All in all I don't believe there's any real benefit in doing this.

PR #78, opened 2017-11-12
Award Checking

SafeteeWoW:
Mostly done. Award checking is done by LOOT_SLOT_CLEARED event.

ML:Award normally returns nil when award success. Now use a callback solution to communicate between ML:Award and PopupOnClickYes. I think this solution is very clean.


SafeteeWoW:
Haven't fully tested this yet, but the code structure should be finalized? comment?

ml_core.lua
```diff
@@ -441,12 +602,12 @@ function RCLootCouncilML:UpdateLootSlots()
...
 			if not self.lootTable[session].awarded and not updatedLootSlot[session] then
-				if item == self.lootTable[session].link then
+				if addon:ItemIsItem(item, self.lootTable[session].link) then
 					if i ~= self.lootTable[session].lootSlot then -- It has changed!
 						addon:DebugLog("lootSlot @session", session, "Was at:",self.lootTable[session].lootSlot, "is now at:", i)
-						self.lootTable[session].lootSlot = i -- update it
-						updatedLootSlot[session] = true
 					end
+					self.lootTable[session].lootSlot = i -- update it
+					updatedLootSlot[session] = true
```
SafeteeWoW: This change is to fix an issue that can cause unawarded session shares the same loot slot. This will cause problem for award checking.

Reproduce:
Session 1 and 2 with identical items on slots 1 and 2. I don't award any of them and close and loot window to update the loot slot. Now both sessions are at slot 1.

ml_core.lua
```diff
@@ -953,12 +1151,21 @@ function RCLootCouncilML.AwardPopupOnShow(frame, data)
...
 		RCLootCouncilML:TrackAndLogLoot(data.winner, data.link, data.responseID, addon.bossName, data.votes, data.gear1, data.gear2,
 		 										  data.reason, data.isToken, data.isTierRoll, data.isRelicRoll, data.note)
 	end
+end
+
+-- Argument to callback: awarded, session, winner, status, data, ...
+function RCLootCouncilML.AwardPopupOnClickYes(frame, data, callback, ...)
+	RCLootCouncilML:Award(data.session, data.winner, data.responseID and addon:GetResponseText(data.responseID, data.isTierRoll, data.isRelicRoll), data.reason,
+		RCLootCouncilML.AwardPopupOnClickYesCallback, data, callback, ...)
+
```
SafeteeWoW: EPGP module will take advantage of the callback argument here

SafeteeWoW:
+ autoAward has no session, so I cant use lootTable in announcement. However, it's really a bad idea to repeat the code to get the item information, so now there is a new function to get item information.
+ FIx a long standing bug that autoAward does not use unambiguate name, so it fails at the later stage.

ml_core.lua
```diff
@@ -433,6 +497,128 @@ function RCLootCouncilML:CanWeLootItem(item, quality)
...
+--@param winner The name of the candidate who we want to give the item to
+--@param callback The callback function that do stuff when this loot attempt success/fail
+--@param ... The additional arguments provided to the callback.
+--@return nil
+function RCLootCouncilML:GiveLoot(slot, winner, callback, ...)
+	if self.lootOpen then
+
+		local entryInQueue = {slot = slot, callback = callback, args = {...}, }
+		entryInQueue.timer = self:ScheduleTimer(OnGiveLootTimeout, LOOT_TIMEOUT, entryInQueue)
+		tinsert(self.lootQueue, entryInQueue)
```
evil-morfar: There isn't really a reason to store it in a table - not that it matters that much.

ml_core.lua
```diff
@@ -493,25 +679,70 @@ function RCLootCouncilML:LootOnClick(button)
...
+	if cause == "loot_not_open" then
+		addon:Print(L["Unable to give out loot without the loot window open."])
+	elseif cause == "timeout" then
+		addon:Print(format(L["Timeout when giving 'item' to 'player'"], item, addon.Ambiguate(winner)), " - ", L["Player is not in this instance or his inventory is full"])
+	elseif cause == "locked" then
+		addon:SessionError("No permission to loot the item at slot "..slot)
+	else
+		local prefix = format(L["Unable to give 'item' to 'player'"], item, addon.Ambiguate(winner)).."  - "
+		if cause == "loot_gone" then
+			addon:Print(prefix, _G.LOOT_GONE) -- "Item already looted."
```
evil-morfar: This is most likely an error on our part (unless the ML has manually awarded the item) so it could probably do with a better explanation?

evil-morfar:
>Haven't fully tested this yet, but the code structure should be finalized. comment?

Seems nice.
We should really use some of the reasons from `:CanGiveLoot()` as responses in the votingFrame (instead of the current "Offline/not installed"). This will break backwards compability though.


> Should we disable award items for awarded session in direct mode? Consider the case where every item is identical.

I'm not sure what you mean. Are you talking about awarding an item that has already been awarded? That should be handled in the voting frame (which I thought it already was) so it isn't an option for the ML.

ml_core.lua
SafeteeWoW: This happens when ML has awarded the loot in the loot window, but then try to award the item in RC

evil-morfar:
There's no need to complicate functions with things they shouldn't handle. I.e. `ML:Award()` should attempt to award items, and while it could deny already awarded items, it's better to just not give the user the option to award already awarded options. This helps to keep the coupling low, which in turn makes for better readability and better debugging.

Locale/enUS.lua
```diff
@@ -93,6 +93,7 @@ L["award"] = true
 L["Award Announcement"] = true
 L["Award for ..."] = true
 L["Award later?"] = true
+L["Award later isn't supported when testing."] = true
```
evil-morfar: I think the previous message was fine

ml_core.lua
```diff
@@ -752,37 +766,64 @@ end
...
+	if self.lootTable[session].lootSlot and self.lootTable[session].bagged then -- For debugging purpose, addon bug if this happens, such values never exist at any time.
+		addon:Session("Session "..session.." has unlooted item in the bag!?")
+		return awardFailed(session, winner, "unlooted_in_bag", callback, ...)
+	end
+	if not winner and not self.lootTable[session].lootSlot then -- For debugging purpose, addon bug if this happens. UI should not allow this.
+		addon:Session("The item of Session "..session.." cant be awarded later because it is not in the loot window.")
+		return awardFailed(session, winner, "bagging_no_loot_slot", callback, ...)
+	end
+
+	if self.lootTable[session].awarded then -- already awarded. Change award
```
evil-morfar: Changing an award should have it's own dedicated function

ml_core.lua
```diff
@@ -752,37 +766,64 @@ end
...
+	if self.lootTable[session].bagged then  -- indirect mode (the item is in a bag)
+		-- Add to the list of awarded items in MLs bags, 
+		local entry = {link = self.lootTable[session].link, winner = winner} -- Note that we have ensured winner is not nil here, by the checkings in the earlier code.
+		self.lootTable[session].awardedInBagsEntry = entry
+		tinsert(self.awardedInBags, entry)
+
+		-- Remove it from lootInBags
+		if type(self.lootTable[session].bagged) == "number" then  -- "number" type indicates the item is awarded later, not added by "/rc add". Items from "/rc add" are not in lootInBags
+			for k, v in ipairs(self.lootInBags) do -- It's a bug to replace this loop by: tremove(self.lootInBags, session _OR_ self.lootTable[session].bagged)
+													-- The reason is similar to why we need to UpdateLootSlots() for direct loot: Index changes.
```
evil-morfar: Except here we don't care about the index, we just need it in case we have multiples of the same item. But great approach anyway - the same should be used for `awardedInBags` after trading to another player in #53.

ml_core.lua
```diff
@@ -794,40 +835,37 @@ function RCLootCouncilML:Award(session, winner, response, reason, callback, ...)
...
 				end
 			end)	
-		else -- Store in bags and award later
-			self:GiveLoot(self.lootTable[session].lootSlot, winner, function(awarded, cause)
+		else -- Store in our bags and award later
+			self:GiveLoot(self.lootTable[session].lootSlot, self.playerName, function(awarded, cause)
 				if awarded then
 					tinsert(self.lootInBags, self.lootTable[session].link) -- and store data
+					self.lootTable[session].lootSlot = nil
+					self.lootTable[session].bagged = true  -- Now the item is bagged and no longer in the loot window.
```
evil-morfar: Doesn't matter - the item is tagged as awarded and should never be able to enter this function again.

ml_core.lua
```diff
@@ -752,37 +766,64 @@ end
...
+	if not winner and not self.lootTable[session].lootSlot then -- For debugging purpose, addon bug if this happens. UI should not allow this.
+		addon:Session("The item of Session "..session.." cant be awarded later because it is not in the loot window.")
+		return awardFailed(session, winner, "bagging_no_loot_slot", callback, ...)
+	end
+
+	if self.lootTable[session].awarded then -- already awarded. Change award 
+		registerAndAnnounceAward(session, winner, response, reason)
+
+		if self.lootTable[session].bagged then
+			self.lootTable[session].awardedInBagsEntry.winner = winner
```
evil-morfar: `awardedInBagsEntry` doesn't matter - we're most likely wiping the lootTable before this could even be used anyway.

ml_core.lua
SafeteeWoW: Now index is no longer used. The table entry is stored instead.

ml_core.lua
SafeteeWoW: Bagged item is not tagged as awarded

ml_core.lua
SafeteeWoW: I'm not spending extra effort to create another function. Changing award in the same function makes sense for me.

Locale/enUS.lua
SafeteeWoW: What do you mean?

ml_core.lua
SafeteeWoW: Have done lots of rewrite since your comment

ml_core.lua
```diff
@@ -462,6 +653,128 @@ function RCLootCouncilML:CanWeLootItem(item, quality)
...
+			if addon:UnitIsUnit(GetMasterLootCandidate(slot, i), winner) then
+				found = true
+				break
+			end
+		end
+		if not found then
+			return false, "not_ml_candidate"
+		end
+
+		-- IDEA: I don't check this now. but is there any way to check if the unit is in the instance?
```
evil-morfar: Not as far as I know. That's why I have the `NOTINRAID` response delivered by the candidate (although that simply checks `IsInInstance()`). That response could be used to check for it, although it'll already be listed as the response.

SafeteeWoW: All people to have RC to make this work. Use commutation to check this does not guarantee a correct result any way. Not going to check this, and this is not important to check.

SafeteeWoW:
+ "/rc add" now set lootTable[session].bagged as false, because it does not add the item to the item list we store.
+ lootInBags and awardedInBags are combined into one table ```baggedItems```

SafeteeWoW:
Several commands need to be added to support the new award later feature and I haven't added yet. I'm bugfixing the current features.

SafeteeWoW:
I should have cleared most of the bugs.

SafeteeWoW:
+ added function ```ML:GetLootTableForTransmit```. I know you said it is weird to have sth added but removed later, but I think it's okay. There can be sth in the loottable only useful for the ML. And the main reason I added that because I added ```lootTable[session].baggedEntry``` and this is used to properly modifying the entry without worrying about the index of the ```addon.baggedItems``` changes. If I store ```baggedIndex``` instead, then I have to make sure no entries are removed from ```addon.baggedItems``` during the session.    ```lootTable[session].baggedEntry``` should not be transmitted.

SafeteeWoW:
I just need to do some locale clean up and more testing right now.

Locale/enUS.lua
```diff
@@ -283,6 +299,7 @@ L["Print Responses"] = true
 L["print_response_desc"] = "Print your response in the chat window"
 L["Protector Token"] = true
 L["Raw lua output. Doesn't work well with date selection."] = true
+L["rclootcouncil_trade_add_item_confirm"] = "RCLootCouncil detects that %s tradable items in your bag are awarded to %s in the award later list. Do you want to add the items to the trade window?"
```
evil-morfar: Has this been tested? Pretty sure it should be %d as it's stored as a number.

ml_core.lua
```diff
@@ -396,25 +649,106 @@ end
...
+						tinsert(tradedItemsInBag, link)
+						break
+					end
+				end
+			end
+
+			if #tradedItemsInBag > 0 then
+				addon:Print(format(L["The following items are removed from the award later list and traded to 'player'"], addon:GetUnitClassColoredName(self.tradeTarget)))
+				addon:Print(table.concat(tradedItemsInBag))
+			end
```
evil-morfar: Isn't this a bit redundant? I mean, they literaly just saw the items being traded.

Modules/votingFrame.lua
```diff
@@ -1158,27 +1173,36 @@ do
 				text = "",
 				notCheckable = true,
 				disabled = true,
-			},	{ -- 6 Change response
+			},{ -- 6 Award later
+				text = L["Award later"],
+				notCheckable = true,
+				disabled = function() 
+					return not lootTable[session] or lootTable[session].bagged or lootTable[session].awarded
```
evil-morfar: This seems wrong. Why shouldn't a bagged item be awarded later? 

Award later would also be enabled for an item that is awarded and not bagged - that should probably be the other way around.

evil-morfar:
I changed these 9b63e8c8312ebdb971bdab4e319138c2d6cfe6f3

evil-morfar:
Can you confirm I covered all the changes of this PR? 4a47718927bb131bf1fcb3e9dc62e6a3d8672ed6

evil-morfar:
@SafeteeWoW

PR #81, opened 2017-11-13
30s default timeout is too short

SafeteeWoW:
+ 30s default timeout is too short and this is very annoying default setting for new RCLootCouncil user. Let's set it to 180s, which is the Blizzard timeout for bonus rolls confirmation.
  + I prefer this setting over disable timeout by default, because this teaches the user a timeout feature exists.

### DIFF
--- a/Modules/options.lua
+++ b/Modules/options.lua
@@ -963,7 +963,7 @@ function addon:OptionsTable()
 											if self.db.profile.timeout then
 												self.db.profile.timeout = false
 											else
-												self.db.profile.timeout = 30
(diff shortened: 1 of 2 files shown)


evil-morfar:
While I'll agree 30 seconds is probably too short, 3 minutes is definitely too long - so long it kinda defeats the purpose of the timer. Your argument is also kinda invalid, as a 30s timer really prompts users to go check out the settings to change it.

45s - 60s is much more reasonable imo.

evil-morfar:
11ac48a44fee716ee8c9d06b02d0f458fd454474

PR #82, opened 2017-11-13
Use FindInTableIf instead of tContains in council check

SafeteeWoW:
```
-- Defined in SharedXML/Util.lua:199
function FindInTableIf(tbl, pred)
	for k, v in pairs(tbl) do
		if (pred(v)) then
			return k, v;
		end
	end

	return nil;
end
```
I don't like to do raw comparison to the name, especially when the operation has vert bad effect if failed.

### DIFF
--- a/core.lua
+++ b/core.lua
@@ -697,7 +697,7 @@ function RCLootCouncil:OnCommReceived(prefix, serializedMsg, distri, sender)
 					end
 
 					-- Check if council is received
-					if not tContains(self.council, self.masterLooter) then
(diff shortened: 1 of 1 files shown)


evil-morfar:
The idea is it's the same function generating the names, thus should be ensured to be the same. I originally created `:UnitIsUnit()` back when they added connected realms where names sometimes had realm name appended, and sometimes not. That was later extended back when I didn't know why different functions returned different capitalizations, which I later "fixed" by semi forcing titlecase. 

I suppose it makes sense to include the check, just in case of something weird happening, say between different locales.

PR #83, opened 2017-11-13
Dont disable award menu for awarded item

SafeteeWoW:
+ Dont disable entire rightclick menu when the item has been awarded.
+ Reaward will be handled in #78


SafeteeWoW:
Still considering the design decision. Close this for now

evil-morfar:
My initial thought was it's easier to just not show the rightclickmenu when awarded, as there's (currently) nothing in there that should be accessible when the item is awarded (See also comment in #78).

I did initially disable every option in the rightclickmenu - I'll revert back to that, as I don't see a better alternative.

PR #86, opened 2017-11-14
Disable sorting in session frame

SafeteeWoW:
Force all columns to be sorted by real row, effectively disable sorting.


evil-morfar:
Might as well do it properly 16347515947903097f34f63da68feab72fac853e.

PR #90, opened 2017-11-18
Add roll button sorts rows by the column "roll"

SafeteeWoW:


evil-morfar:
Response already has roll as its sortnext. Imo response should always be sorted first, no matter what the rolls are.

PR #91, opened 2017-11-18
Award sets the response to Awarded

SafeteeWoW:
+ When the item is awarded, sets the response of the awardee to "Awarded" to all duplicate items, so I don't need to remember who have been awarded one of the duplicate items, so I never mistakenly award two identical items to the same people
+ This breaks backward compatibility though


evil-morfar:
The only issue with this is it doesn't account for reawards.

PR #92, opened 2017-11-19
Reannounce and roll improvement

SafeteeWoW:
+ Add an button to reannounce this item to everyone
+ Reroll now also does autopass
+ Reroll no longer only reroll unawarded items.
   + I think it doesn't make sense to reroll only unawarded items, because the text didn't mention "unawarded"
+ Cleaned up some duplicate code
+ Manual roll feature has been abandoned.
+ New roll feature: Reannounce to everyone and request rolls. When the user responses in the lootFrame, they automatically does a system roll and announce it in chat and report the result to the ML. (Only do system roll if the response is not "AUTOPASS", "PASS", "TIMEOUT", "DISABLED", "NOTINRAID")
  + As a new feature, this feature does not support backward compatibility.
+ There is also another button to request roll from anyone whose roll is the same as the selected candidate, which helps to resolve tiebreakers quickly.
+ 
wowscrnshot_111917_033341
ML can choose to if he wants to request rolls from raid members for each row in the session frame.
+ Auto add rolls won't be active on session that requests rolls
+ This PR should be finished if I don't find any bug


SafeteeWoW:
+ Pros of this feature
  + Transparent rolls
  + Make the roll more exciting compared to every roll is done in the voting frame.
  + This is much better than the previous manual roll, because this is more automatic now.
+ Cons:
  + Have to print message which is similar to Blizzard group loot rolls system message to differ the roll to each item. This makes the chat crowded, but I have to in order to implement this feature. But it shouldn't be a big deal, as people usually rolls only on several items

evil-morfar:
Imo rerolls shouldn't be autopassed - it's the only way to "force" a player to roll on a item. That's basically why you added "reroll items the player can use"?

I don't think the button for rolls on each session is necessary - if they really want a roll on specfic session they can do it from the rightclick menu.

I'm also not quite sure if I like the amount of buttons in the rightclickmenu - should probably consider nesting more of them.

Other than that I like it.

evil-morfar:
Also reroll shouldn't send out awarded items. It doesn't matter if it isn't mentioned - you wouldn't expect already awarded items to be rerolled by default - if you want that you should specificly only reroll that item.

SafeteeWoW:
> Imo rerolls shouldn't be autopassed - it's the only way to "force" a player to roll on a item. That's basically why you added "reroll items the player can use"?

Ok, but rerolls broadcasted to everyone will have autopass active.

> Also reroll shouldn't send out awarded items. It doesn't matter if it isn't mentioned - you wouldn't expect already awarded items to be rerolled by default - if you want that you should specificly only reroll that item.

Now RCLootCouncil supports to change award. No need to only reroll unawarded items because I can change the award to someone else.

> I don't think the button for rolls on each session is necessary - if they really want a roll on specfic session they can do it from the rightclick menu.

I think it's needed. Otherwise if the ML wants people to roll on an item, everyone has to click lootFrame on that item once, then ML requests to roll, everyone has to click lootFrame again. I really don't like this redundancy.  Sometime it's obvious to see an item is bad and only worth to be rolled, in the session frame. Sometime ML decides to roll only when he sees everyone's response.

SafeteeWoW:
wowscrnshot_112017_035034

SafeteeWoW:
I have done nesting the menu. How is this?

SafeteeWoW:
wowscrnshot_112017_035034

It looks like this now. Btw, I have tried the minimize the duplicate code

SafeteeWoW:
I have finished this PR. Haven't found any other bug.

SafeteeWoW:
I have tested and finished this PR

evil-morfar:
#### Reroll
I'm fine with autopassing on "Reannounce all items", I just read it like all rerolls would be autopassed. As long as "Reannounce this item" isn't autopassed, then it's fine. 

The base usecase hasn't changed - no matter which features are implemented. By default you wouldn't even start to award an item if you knew responses are missing. The usecase is someone timeouted/something and we want to reannounce the items to him. Since he autopassed/couldn't use the first two items, we went ahead and awarded them - but we need his response for the remaining, and only the remaining, non-awarded items.
The reaward thing is mainly for fuckups, and is not included in any basic reannounce usecases.

#### Rolls
The reason I don't like the roll buttons on the session frame (besides the extra complexity it adds) is I don't think it's worth putting such a minor feature as rolls on a front row seat. The intent with rolls is to add some randomness when it's needed - which it rarely should be. I understand some people wants to always show these rolls (which they already can) and thus not really use the addon as a loot council addon. But at its core it's intended for use as a loot council tool, not a "easy random roll machine".

You're arguing those buttons needs to be there because the roll is tied to the lootframe - in that case maybe the rolls shouldn't be tied directly to the lootframe. I think your former "manual roll" was fine. If you want to make it more automatic, why not just do a ```print("You rolled xxx")``` in the candidates chat? Or maybe do a popup if you really want to involve the users.

#### Rightclickmenu
I see no reason why you'd like to have class, rank or role as a seperate entry to reannounces. Even response and roll doesn't really seem necessary. Imo it's all just clutter that'll probably never be used.

SafeteeWoW:
###### Rightclickmenu
I can remove class, rank, or role, but response and roll are definitely useful. Roll is used to run tiebreakers in case roll is the same. Response is used in case you want to request rolls from people who clicked "Need". In those scenarios, these options are very very useful.

evil-morfar:
##### Purpose of the roll feature
1. I totally undestand. But with that argument, your original "/roll" thing makes more sense.
2. You could do that numerous ways. Again, your original idea already did that.
3. That's a matter of perspective. You could even do a ```print()``` with the right color to make it indifferent from the server roll. The actual code used for the roll is not different.

##### Rightclickmenu
Except in my opinion rolls should not be related to reroll/response. I.e. requesting a new roll (1-100) should not prompt a new response request. 

I do see some potential in requesting a reroll for a certain response (albeit that would probably be very rare). Likewise, it's a fine way to handle ties (but I prefer your addition that makes ties impossible) - but it doesn't relate to having people seeing the loot frame again. I honestly think that people would have requested it if they thought they needed more "roll" stuff, but all I've ever seen was a request to extend the current "ML rolls for you, without you seeing your roll" thing to be automatic. Tying rolls to responses is bad, as it's literally two different things, and we don't need to show the loot frame to make a /roll.

##### Rolls button on session
It might be important _sometimes_, but as a general rule, it's not. Even if you really want to have rolls every time (there's an option for that), that doesn't mean you should choose per session. Everything or nothing (and then a few, (i.e. "request rolls" if you change your mind)) is enough.

The extra width in itself is complexity enough. Not to mention mousing over an empty frame just to see a faded dice button appear (btw, I think that's a nice touch) - I don't imagine people knowing what that means. Eitherway, it's complexity for the majority that probably won't use it.

>>* Only know what your own roll is not useful. You need to know everyone's roll. Therefore, print in chat is required.
>>* Again, it's different feeling to roll for yourself instead of ML rolling for you. No matter if ML's roll result is printed to you.

Now we're back to your orignal roll feature. 

>>* The main reason that I added rolls button on session is that I don't like redundancy. I don't want to be forced to let raid members response once, then request rolls from raid members, then they need to click again, even if I want to request rolls at the beginning. This does not feel good.

But yet you're still offering that. And that's also excactly why lootframe and rolls should be different things.

SafeteeWoW:
> But yet you're still offering that. And that's also excactly why lootframe and rolls should be different things.
See Blizzard Group Loot Frame for example. There is reason that roll is binded with NEED/GREEE/DISENCHANT/PASS.

SafeteeWoW:
> You don't do that in a loot council. You only need the roll after the response can't determine the winner.
There are multiple cases when the response cant determine the winner. Some people roll minor upgrade for main spec. Some roll for offspec. Some roll for transmog.

evil-morfar:
>>You don't do that in a loot council. You only need the roll after the response can't determine the winner.

Keyword: AFTER. You don't know that beforehand, in which case it's nice to add/ask for rolls, which shouldn't include a new response request.
People that changes their mind is not supported in a loot council. You state whether you want the item or not, anything after that is up to the council.

>>There is a tooltip telling what it means when mouseovering the dice.

Except it's a new concept in RCLootCouncil, that's not usually used in a loot council.

>>It's not the time to predict that "majority" won't use it.

Maybe not, but I prefer having data to support a statement, and when I don't, I prefer to go with my own experience - and that tells me that this has never been requested, nor does anyone I've talked to wanted to actually use it.

SafeteeWoW:
> Maybe not, but I prefer having data to support a statement, and when I don't, I prefer to go with my own experience - and that tells me that this has never been requested, nor does anyone I've talked to wanted to actually use it.

"A lot of times, people don't know what they want until you show it to them."

SafeteeWoW:
> People that changes their mind is not supported in a loot council. You state whether you want the item or not, anything after that is up to the council.
The different conditions given by the ML(roll/not roll) causes the decision. If the guild runs DKP/EPGP, item being rolled has no "price", it affects the decision so much.

SafeteeWoW:
I will make a change.

SafeteeWoW:
I have applied your suggestions:
+ All changes to session frame have been reverted.
+ request roll only request rolls. No response.
   + Note button is hidden
   + Only dice and pass icon are shown
     + I don't use a button with the text "Roll" and "Pass" because I want these frame looks distinct.

Before dice is clicked
wowscrnshot_112517_002123
After dice is clicked. The frame will be hided after 1s
wowscrnshot_112517_002126

The current icon size is the maximum size to be correctly shown in zhCN locale

evil-morfar:
Steve Jobs might have been somewhat inspirational, but in this case I prefer "I cannot give you the formula for success, but I can give you the formula for failure--It is: Try to please everybody."

You show the number of items because people likes to know it, and showing it saves them the time to go look at the boss.
Guilds running epgp would probably always have rolls enabled thus not needing to decide per item.

evil-morfar:
I actually really like that!
But what about requesting rolls for all sessions before starting it?

SafeteeWoW:
This requires another design, no one want to click twice for single item

SafeteeWoW:
Not sure how to do it and I prefer way to do this is still the session frame changes I did a while ago. 
If the setting is not in the session frame, but in the option panel, always requesting rolls from everyone for all items cause spamming problem.

PR #93, opened 2017-11-19
loot frame should start over when receives loot table

SafeteeWoW:
+ Fix a bug that voting frame and loot frame could fail if lootTable is received when they are shown.
+ Fixed by disable them first when the lootTable is received.
+ I comment out ```self.frame:SetParent(nil) self.frame = nil``` in onDisable of votingFrame. Otherwise there is memory leak every time the voting frame is disabled
+ I have tested and monitored the memory usage to ensure there is no significant memory leak.

### DIFF
--- a/core.lua
+++ b/core.lua
@@ -733,7 +733,9 @@ function RCLootCouncil:OnCommReceived(prefix, serializedMsg, distri, sender)
 					end
 
 					-- Hand the lootTable to the votingFrame
+					self:GetActiveModule("votingframe"):Disable() -- Disable first, in case voting frame receives the lootTable when it is shown. (It can happen if ML reloads.)
 					if self.isCouncil or self.mldb.observe then
+						self:CallModule("votingframe")
 						self:GetActiveModule("votingframe"):ReceiveLootTable(lootTable)
 					end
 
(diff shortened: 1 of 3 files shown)


SafeteeWoW:
+ Pros of this feature
  + Transparent rolls
  + Make the roll more exciting compared to every roll is done in the voting frame.
  + This is much better than the previous manual roll, because this is more automatic now.
+ Cons:
  + Have to print message which is similar to Blizzard group loot rolls system message to differ the roll to each item.

core.lua
```diff
@@ -775,9 +778,7 @@ function RCLootCouncil:OnCommReceived(prefix, serializedMsg, distri, sender)
 				self.isCouncil = self:IsCouncil(self.playerName)
 
 				-- prepare the voting frame for the right people
-				if self.isCouncil or self.mldb.observe then
-					self:CallModule("votingframe")
-				else
+				if not (self.isCouncil or self.mldb.observe) then
```
evil-morfar: I don't think this is good - voting frame needs to be enabled to receive comms related to it - it might receive those at the same time as loottable.

evil-morfar:
I don't understand why they need to be disabled? This has never been a problem before.

SafeteeWoW:
+ Fix an issue if ML reloads during session and start session again, before other raid members haven't finished clicking lootFrame.
  + Reproduce: Play two characters and put them in a group. Group leader does /rc test and reload immediately when see the voting frame, then /rc test again. The display of the loot frame of the other raid member is unintended.
+ Fix an unintended code that it's possible that :Trash() fails to remove the entry from self.entries.
  + Before the fix, it's possible that self.entry has some entry left, after the LootFrame is disabled. 
  + To make tremove(self.entries, entry.position) works, :Update() must be run after every trash, but you didn't run :Update() after each :Trash() in :OnDisable(). I change tremove to tDeleteItem. Now entry.position is only used in debugging code. I don't think it's a good idea to maintain such variable.

PR #97, opened 2017-11-24
Sync modules

SafeteeWoW:
+ The setting of modules can now be synchronized through "/rc sync"
+ Modules need to run ```registerDefaults``` after sync, otherwise, it does not work.
+ profile should be wiped first, otherwise it is not guaranteed that our setting is the same as the sender

### DIFF
--- a/Modules/options.lua
+++ b/Modules/options.lua
@@ -782,7 +782,7 @@ function addon:OptionsTable()
 									outputDesc = {
 										order = 2,
 										fontSize = "medium",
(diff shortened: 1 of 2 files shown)


evil-morfar:
The intent with the global sync handler is for modules to add it themself. RCLootCouncil cannot know where all of a module's settings is located, nor assume modules actually store their stuff in the right place.

While you're right about the current profile should be wiped, it's not that simple. For example the bonus roll added to ```db.responses``` would be removed if a user using EU receives settings from someone that doesn't. 

It's also only possible to sync settings that the sender actually has - if we have another set of modules installed we don't expect the sync to change any of those settings.

PR #99, opened 2017-11-24
module version check

SafeteeWoW:


evil-morfar:
This would potentially be very spammy. My initial thought was to have modules depending on everyone having it installed to handle that in their own code. For general upgrade information I intended the version check frame to just color the text red for old versions.

While RCLootCouncil could easily handle a simple version check, I think it should be up to the modules to determine how they handle outdated versions. 

Either way, spamming is out of the question.

SafeteeWoW:
Not sure why it can be spamming.
It's only possible to have one line print per module per login. How can it be spamming?

evil-morfar:
Ahh I missed ```self.moduleVerCheckDisplayed``` when first looking over it. Eitherway, my other points still stand.

SafeteeWoW:
+ How will modules handle outdated version different? For EU, GroupGear, EPGP, all they do when get a outdated version should be printing the new version. Is there any difference?
+ If you intends to let the modules to handle the version check themselves, why do you collect the version information of modules in the main addon?

evil-morfar:
>>How will modules handle outdated version different? For EU, GroupGear, EPGP, all they do when get a outdated version should be printing the new version. Is there any difference?

EU and GroupGear doesn't care about other people's version by design. But for example if EU at some point do, I would very much like to a include a message, stating that this feature can't be computed from x user as their version is outdated.

Again, RCLootCouncil can easily handle the general "outofdate" check, but shouldn't do prints, as the modules might have their own statement to add.

>>If you intends to let the modules to handle the version check themselves, why do you collect the version information of modules in the main addon?

To show which modules a user has installed. The version number is just an addition to that. And that displaying of the module is where any module version check should be displayed, as there's not really another good way to show it.

evil-morfar:
Didn't see your edited post with my last comment.
>>Version check(New version detection) don't need to have everyone installed the module.

No, and that has never been an issue.

>>It's only possible to have one line print per module per login. It does not spam.

True, see comment above.

>>It only checks the version of the modules you have INSTALLED. If the user only installs RCLooCouncil, but not EPGP/EU/GroupGear, he won't get upgrade message for EPGP/EU/GroupGear.

Of course? I don't see what this has to do with anything?

>>I see no reason to write duplicate code for version checking in the module, when it is already there in the main addon

And that's why you can easily fetch ```verTest``` comms and decide what to do. You can even use the ```:VersionCompare```, which would reduce the amount of code needed to ~6 if you want to do a "outdated" print.

>>While I agree that syncing should be added by the module itself because database management is different for each module, outdated version detection is really the same for any module.

Not sure how sync enters this conversation, but I agree. But again, if you want to do something different than a "outdated" print, including a core version checking, would produce multiple prints to the user, with the RCLootCouncil "this module is outdated" not really being relevant. Overall, I think modules should handle it themselves.

SafeteeWoW:
> EU and GroupGear doesn't care about other people's version by design. 
True, but why not new version detection? Some addons don't even have commucation feature, but still has new version detection, like GTFO.

evil-morfar:
I'm inclined to agree (and always have been btw) - I just don't like the idea of RCLootCouncil printing stuff to the user, if the module had it's own plan to print something.

The chat scenario:

* RCLootCouncil: Module "ExtraUtilities" 0.7 is outofdate - newest version is 0.8.
* ExtraUtilities: Your version is outdated, please upgrade.


doesn't suit me.

SafeteeWoW:
Why would module author print "Your version is outdated, please upgrade." when he should know core has done it?

evil-morfar:
Vertest should probably also be included though

PR #106, opened 2017-11-28
Fix loot table sorting

SafeteeWoW:
+ Complete forget to edit this when localized subType is removed.
+ The sort order is the same as before the fix. (Except weapon is sorted after relic now, but there is no weapon in this expansion, so doesn't matter.)
+ This is the reason why we should clear our saved variables when testing.


evil-morfar:
As a side note - I always test release versions with fresh SV - I just did that the day before when I thought everything was in, and didn't do it again when the localizedSubtypes was cleared. Morale of the story; last minute changes are really bad, and should be avoided.

PR #107, opened 2017-11-28
Attempt to fix mldb

SafeteeWoW:
+ When receiving MLdb, we should check if it is from the ML.
  + I suspect this is the reason of the recent reports of MLdb is not being received because some non-ML enables ML module and send MLdb
  + Even if this is not the reason, we should always check it, in case someone's version messes up
  + Although you claimed long ago " NOTE: 2.1.7 - While a check for this does make sense, I'm just really tired of mldb problems, and noone should really be able to send it without being ML in the first place. So just accept it as is.", but this is really never the reason why the mldb is not received. Not checking this only creates problem when any non-ML somehow sends mldb. We should always check if the ml commands are from the ml.
+ Also clear mldb when ml changes.
   + Then, we'll only have either no mldb, or mldb of the current ML. We'll never have mldb of someone else. If somehow mldb is not received, this is be detected in OnCommReceived and mldb should be resended.
   + I have considered to also clear ```self.candidate```, but that's also a part of data sent by ml.

### DIFF
--- a/core.lua
+++ b/core.lua
@@ -1771,6 +1771,7 @@ function RCLootCouncil:NewMLCheck()
 		-- At this point we know the ML has changed, so we can wipe the council
 		self:Debug("Resetting council as we have a new ML!")
 		self.council = {}
+		self.mldb = {}
 		self.isCouncil = false
 		self:Debug("MasterLooter = ", self.masterLooter)
 		-- Check to see if we have recieved mldb within 15 secs, otherwise request it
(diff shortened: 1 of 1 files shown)


evil-morfar:
The problem (why I removed the check) was that the ML could send the mldb before users registered him as ML. I suppose now that the groupleader is always registered as ML, that won't be an issue (except when doing a /reload in a raid). I still don't see how random people can send out mldb?

Clearing the mldb on a new ML is probably a good idea, as we're guaranteed to receive it within 15 secs (should probably also be lowered).

Also, there's no reason to clear candidates - it's really just an indentifier list with some extra info, and it will be updated whenever needed.

evil-morfar:
Testing with starter accounts (no ML) always gets the groupleader as ML before mldb is received both when joining and when /reloading. I suppose the this will be the case even if there's a delay on receiving ML status, although I would really like it confirmed with ML before including such a potential breaking change.

evil-morfar:
Tested with ML in a small group - seems to be working.

PR #110, opened 2017-11-29
Remove function IsFiltering, replace it by a one-line code.

SafeteeWoW:
There are several issues in ```IsFiltering``` function.
+ Unneeded redundent code, make it harder to expand the filter feature.
+ This function has issue when the saved variable ```db.modules["RCVotingFrame"].filters``` contains garbage. It bugged if I active the filter for a button, but later removed the button from responses..
+ Replace this function by one line solution: ```#self.frame.st.filtered < #self.frame.st.data```
  + Although it does not change the color of filter button when a filter is active but no rows are filtered, I think it does not matter.

### DIFF
--- a/Modules/votingFrame.lua
+++ b/Modules/votingFrame.lua
@@ -398,34 +398,6 @@ end
 --	Visuals
 -- @section Visuals
 ------------------------------------------------------------------
--- Returns true if a filter is set for this session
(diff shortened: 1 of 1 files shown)


evil-morfar:
Brilliant solution

PR #111, opened 2017-11-29
Filter by guild rank

SafeteeWoW:
+ This could be useful for guild running split raid.
+ Can you write an introduction of filter feature on the Curseforge page?

---
+ If you think it's useful, merge it, otherwise nvm


evil-morfar:
Fine feature - as long as the actual ranks are nested, so that we don't potentially have the filter list populated with a huge number of guildranks. Also, what's the "other" rank?

I think the filtering is pretty self explanatory and doesn't need an introduction.

SafeteeWoW:
Menu has been nested

PR #115, opened 2017-11-30
Print sth when request roll or reannounce.

SafeteeWoW:
+ Print sth when click buttons or click confirmation dialog when request rolls or reannounce.
  + Request roll feature is lack of visual indication that the button is clicked successfully.


evil-morfar:
I'm not sure about this - clicking a button (potentially even a getting a popup with a second button to click) seems like confirmation enough.

SafeteeWoW:
Click request rolls-> nothing happen immediately if no rolls are currently set -> So I don't know if I clicked it -> Need to print

PR #117, opened 2017-12-01
Add sigil of the dark titan to ignoredItems

SafeteeWoW:
+ I think add default ignored items should no longer be necessary in the future.
   + RC should only loots equipable+relic+mount+tiers. We actually don't need "loot everything" option. Dont think there is any other item that can be master looted.
   + Tiers can be recognized by parsing tooltip of "Class: xxx", or using ```GetItemSpell```
   + The easier way is to check the GetMasterLootCandidate, if there is no other person except ML, dont add it to the session.


evil-morfar:
I agree right now it doesn't seem like there's anything else to loot. But what about if toys are added? Profession stuff maybe? But yeah, it would probably be easier without the ignore items, but seeing as this is the last raid in Legion, we shall have to see if things changes in the next xpac.

PR #121, opened 2017-12-03
Autopass Trinket

SafeteeWoW:
As requested by Ticket [314](https://wow.curseforge.com/projects/rclootcouncil/issues/314), I made this feature.

+The idea is, whenever you enter an instance, we fetch all trinket data in this instance and difficulty from Encounter Journal. (When logged in and not in any instance, fetch info of Argus normal for testing purpose).
If the trinket is in EJ for all classes, but not for our class, autopass it.

+ Just need to more testing to ensure there is no taint and no side effect that alters the functionality of Blizzard Encounter Journal.~~
  + I don't find any tainting caused by ```_G.EncounterJournal:UnregisterEvent("EJ_DIFFICULTY_UPDATE")```. Hope it does not. Have to write this because otherwise if the instance displayed in the EJ interface is not the current instance, my EJ_SelectInstance command will be overriden by commands from EJ interface.
   + In order to fetch the data, I have to change the EJ filter. I did manage to set it and then restore it to ensure the function has no side effect that affects the EJ interface.
+ Added command "/rc trinkettest" or "/rc ttest".

---

+ This feature can be easily done by hardcode the trinket, and much more reliable. But the hardcode will be exported programming from Encounter Journal.
+ But considering some people are super slow to update the addon. I am not sure.


SafeteeWoW:
This version has some issues which affects Blizzard Encounter Journal. Not gonna use this one.

PR #128, opened 2017-12-05
Recache in the next frame instead of after 1s.

SafeteeWoW:


evil-morfar:
027f99efbd4ff239513885cbe544c5a859a4a154

PR #136, opened 2017-12-08
Response send item string instead of item link

SafeteeWoW:
+ We don't need item link which includes item name in the response. Only item string is needed.
+ This saves lots of space.
+ Backward compatible.


evil-morfar:
We do actually need it currently. The ML needs it for the loot history. It might be worth considering though, as the ML will have to upgrade anyway - we just need some extra code to request it either before sending out history (easiest, but most space wasting) or when received (would require everyone to upgrade).

Also with my latest changes, `SendResponse` only sends out player's gear with "NOTINRAID" responses.

PR #122, opened 2017-12-03
Autopass trinket Alternate solution: Hardcode trinket classes

SafeteeWoW:
See Also PR #121 

+ Add developer command:
   + /rc exporttrinketdata to export the class info of trinkets to a window. Copy and paste this code to Util/trinketData.lua
+ Trinket data is hardcoded in Util/TrinketData.lua
+ Add command "/rc trinkettest" or "/rc ttest" for testing.
+ All 26 Antorus trinkets are added to the test.
+ The exports includes all trinkets in all difficulties in all instances in the EJ of the current expansion.
  + No need to hardcode trinkets from earlier expansion. Old trinkets equals to gold for everyone.
+ There is an option ```autoPassTrinket``` to toggle this on/off.
  + Maybe it's possible that people use trinket not for their spec, even if it's half effectiveness.

PR #137, opened 2017-12-09
Limit update rate to 1 frame.

SafeteeWoW:
+ After the most recent Curseforge report, I realize 0.2s update interval is not good, especially when switching session. Therefore, the update rate is changed to every frame.


evil-morfar:
The idea behind 0.2 sec was to not have too many Updates in a short period of time. I don't think several responses/lootAcks comes in within a single frame, so this change basically reverts the Update function back to when there wasn't a check.

Also, why should the 0.2s be bad?

SafeteeWoW:
With 0.2s delay, there is a visual delay when switching session with filter enabled? Can you fix that?

SafeteeWoW:
And can you have a look at #122? Since the major bugs have been cleared, can you merge that in the next version?

evil-morfar:
Yes, that's pretty easy. 254471ab38a58169837d6abd632b2e48477eafd2

There's no reason to make the same comment multiple places.

PR #142, opened 2017-12-21
Loot frame show compare item

SafeteeWoW:
+ I know I can press Shift to toggle current equipped gear in the tooltip, but I highly doubt many people knows it, because I only knows this until yesterday.
+ I think it makes sense to always show gear comparison, similar reason why the gear comparison is always shown in character frame when choosing which gear to swap.
+ If you think it's too messy to show equipped gear when mouseovering, then at least the permanently shown tooltip should always show equipped gear, and you can add an option if the equipped gear should always be shown when mouseovering


evil-morfar:
I'm not sure how you didn't know that - everywhere else in the game you can hold shift to compare whenver you see a tooltip.

It somewhat makes sense to do, as users most likely wants to compare the item to their own - I'll have to get a feel for it in game though, which probably won't be until after christmas. Expanding it to handle tier tokens is a good idea though.

PR #144, opened 2018-01-02
Show item bonus text

SafeteeWoW:
+ The item bonus text(socket/leech/avoidance/speed/indestructible) is now shown in session frame, loot frame and the voting frame.
wowscrnshot_010118_204501
wowscrnshot_010118_204603
+ The item state text in the voting frame (green text such as "Mythic Warforged") is removed. One reason is that the item state text is too far right when item bonus is shown with the ilvl. The other reason is that item state is not sth really needed. Everyone can know whether an item is warforged/titanforged according to its ilvl.
+ I have used this PR as ML for two weeks without a problem.

### DIFF
--- a/Locale/enUS.lua
+++ b/Locale/enUS.lua
@@ -382,6 +382,7 @@ L["Skins"] = true
 L["skins_description"] = "Select one of the default skins or create your own. Note this is purely aesthetic. Open the version checker to see the results right away ('/rc version')."
 L["Something went wrong :'("] = true
 L["Something went wrong during syncing, please try again."] = true
+L["Socket"] = true
 L["Sort Items"] = true
 L["sort_items_desc"] = "Sort sessions by item type and item level."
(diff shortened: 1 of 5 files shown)


evil-morfar:
While one could argue "Heroic" and "Mythic" tags are redundant, I very much like the Warforged/Titanforged strings, and consider it a feature that shouldn't be removed. I think there's several alternatives to make room for the new indicators, including resizing and color changes that could make it fit nicely. I'll play around with it and see what I like best. Feel free to give a suggestion though.

I'm also not sure if I like the position in the session frame.

evil-morfar:
I think this is a bit cleaner, but your thoughts are welcome.

image
image
And probably one of the worst cases:
image

LootFrame:
image

Session setup:
image


I'm considering not showing it in the session frame, as I don't think it looks particularly good, and I can't find a way to make it look good.

PR #145, opened 2018-01-06
Show detailed trinket category

SafeteeWoW:
+ Show the trinket category according to its loot spec on voting frame and loot frame. For example, Apocalypse Drive is now shown as "Trinket (Tank)" instead of "Trinket"
+ Such description is determined by the loot spec of the trinket in the Encounter Journal. I have summarized all categories of all trinkets in EJ according to its loot spec. You can check it in _G.RCTrinketCategories table. By the help of Blizzard GlobalStrings, no localization entry is added.
+ To do this, the loot spec data of all trinkets in EJ are exported. See comment in the code for more information. ~~~Some tricks are used to deal with 64bit integer.~~~ Because WoW does not offer full support to 64bit integer, and I am worrying that overflow would occur if the user uses 32bit OS, I decide to use string type to store 64bit number instead of number type.
+ Multiple trinkets are now sorted by its categories.
+ Bugfix: Fix an issue Eye of Shatug is not correctly autopassed, because the item id of the actually item drops does not match the item id in the Encounter Journal.
+ I have tested this PR. You can do the test, for example, by "/rc trinkettest SHAMAN"

wowscrnshot_010618_043259
  
---
+ Although at the end I decide to store the data as string instead of 64bit integer, here is my discovery for the 64bit integer support. (There is no 64bit number in this PR. It is stored as string)
+ My OS is Windows10 64bit.
+ Code to confirm WoW can handle 64 bit integer. No print if test passes. Lots of print if the value of exp is changed to 65. The result shows WoW is capable to store and do integer math on 64bit integer. Integer with more bits are not supported.
```
for k = 1, 10000 do -- Repeat for many times
    local exp = 64
    local diff = math.random(1, 2^31 - 1)
    local a = 2^exp - diff -- Generate a random 64 bit number
    local b = 0
    for i = 63, 0, - 1 do -- Test precision
        local k = math.floor(a / 2^i) % 2
        b = b + k * 2^i
    end
    
    if a ~= b or a ~= math.floor(a) or a ~= math.ceil(a) then
        print("Inprecise", a, b, exp, diff)
    end
end
```
+ bitlib in WoW (bit.band, bit.bor, etc) does not support 64bit integers. Wrong result is operand is greater than 2^32-1
+ tonumber and format function does not support 64bit integers.
+ 64bit integer constants is not allowed in source code. For example, 0x1234567890abcd is not allowed in source code, but it is legit to write it as 0x1234*2^32+0x567890abcd
+ The test code above passes when I run WoW in 32bit mode. This can be done in 64bit OS by renaming Wow-64.exe to sth else and then run Wow.exe. Then on the password screen, you will see "Version 7.3.2 (25549)  (Release x86)" instead of "Version 7.3.2 (25549) (Release x64)" on the bottom left corner.
+ I have no idea what will happen when WoW runs on Mac or 32bit Windows.


evil-morfar:
I suppose WoW only emulates bigger numbers using 32 bit ints - I remember they had to make Garrosh Hellscream regen health several times, as his health pool exceeded their number cap.

A very fine solution though!

PR #153, opened 2018-02-14
Autopass Owned Items

SafeteeWoW:
This PR adds a new autopass option which autopasses owned items. (Enabled by default)
This feature autopasses item when all following conditions are met:
1. You have collected the appearance of the item or the item doesn't have an appearance.
2. You have collected an equal or better version of the item. That means, You has an item equipped or in your bag with the same item id and the item level and any single stat of your owned item is not worse than the item being rolled, including secondary stats, #sockets, leech, etc. (Enchant and gems on the item are ignored before comparison.)
3. The item is equippable (not set token, not artifact relic). Because non-equippable items are usually not permanent and are consumed on use.

TODO:
1. Localization text of this PR needs to be adjusted. I prefer you to adjust it if you accept this PR.
2. Currently, this PR does not check your item in the bank or void storage. This is not hard to do, but in order to do this, I would have to store bank item info in saved variable whenever you open the bank because Bank API doesn't return if you aren't at the bank. I'm not sure if you'll like this.


core.lua
```diff
@@ -1598,6 +1604,80 @@ function RCLootCouncil:GetItemClassesAllowedFlag(item)
...
+	if not GetItemInfo(item) then return end
+	tooltipForParsing:SetOwner(UIParent, "ANCHOR_NONE") -- This lines clear the current content of tooltip and set its position off-screen
+	tooltipForParsing:SetHyperlink(item) -- Set the tooltip content and show it, should hide the tooltip before function ends
+
+	for i = 1, tooltipForParsing:NumLines() or 0 do
+		local line = getglobal(tooltipForParsing:GetName()..'TextLeft' .. i)
+		if line and line.GetText then
+			local text = line:GetText() or ""
+			if text == TRANSMOGRIFY_TOOLTIP_APPEARANCE_UNKNOWN then
+				return false
```
evil-morfar: Are there any implications of not calling `tooltip:Hide()`?

core.lua
```diff
@@ -1598,6 +1604,80 @@ function RCLootCouncil:GetItemClassesAllowedFlag(item)
...
+
+		return true
+	end
+end
+
+function RCLootCouncil:OwnEqualOrBetterItem(item)
+	if not item or not GetItemInfo(item) then return end
+
+	for i=INVSLOT_FIRST_EQUIPPED, INVSLOT_LAST_EQUIPPED do
+		local link = GetInventoryItemLink("player", i)
```
evil-morfar: Consider using `GetInventoryItemLink("player", GetInventorySlotInfo(self.INVTYPE_Slots[select(9, GetItemInfo(item))]))` instead of checking every single item.

core.lua
SafeteeWoW: no. I just forgot to do that.

evil-morfar:
A fine idea. How extensively have you tested it? I'm especially nervous about the implications of adding a new response type.

I'll probably want to change some of text, no biggie though.

I don't think it's worth checking the bank.

core.lua
SafeteeWoW: sure

SafeteeWoW:
I did some basic test when I pushed this PR, but I am not available to do any extensively test or make any major changes right now. You can add your reviewed stuff if you want to merge this PR.

evil-morfar:
Roger that. I'm thinking it'll go in v2.8 with the other new stuff.

SafeteeWoW:
Another "tiny" problem:
patterns and replacment in ```GetItemLinkWithoutEnchant``` should probably excludes "|H" in case input is item string instead of item link

SafeteeWoW:
Another small feature suggestion:
Autopass owned mount

SafeteeWoW:
Above suggestion is done.

SafeteeWoW:
I have made another improvement that I have loose the item comparison requirement that it does not always require the two items to have the same id.

If the item being compared is a stat stick (equippable item that not set piece, no on-use effect, no on-equip effect), then we compare our item to it if they are similar (Same type/subtype/equipLoc).

This change has potential to give false autopass. Many tests are needed.

SafeteeWoW:
TODO: Fix dual-wield weapon

SafeteeWoW:
This PR won't be ready for v2.8. After more consideration, I have found that this feature is more complicated than I imagined before, especially for weapons. Many stuffs needs to be considered to avoid false autopass. Meanwhile, I really dislike the current code structure to handle the item information. Current implementation makes the code very redundant, slow and easy to make mistake.

Let me summarize when the item will be autopassed by this PR. I haven't implement the entire definition yet. ___***IMPORTANT: All following ignores the item gems and enchants both item being rolled and item being checks***___

1. The item has no appearance or the item appearance has been collected. 
2. If the item rolled is an one-hand weapon not specifying main or off-hand, the player needs to have 2 equal or better items equipped or in the bags. Otherwise, only need 1 equal or better item.

The definition of equal or better item equipped or in the bag:
1. The item being checked must be usable by the player. (TODO: Not implemented yet)
2. The item being checked must not be tradable. (TODO: Not implemented yet)
3. The ilvl of the item being rolled is not greater than the item being checked.
4. Every single stat (i.e., secondary stats, #sockets, etc) as returned by ```GetItemInfo``` of the item being rolled is not greater than the item being checked.
5. The item being rolled and checked are similar.

The definition of similar items.
+ If two items have the same item ID, they must be similar. Skip the following checks.
1. Two items have the same ```equipLoc``` as returned by ```GetItemInfo```. Special case: ```INVTYPE_ROBE``` and ```INVTYPE_CHEST``` are similar. Note that, this does NOT check Axe/Sword/Dagger/Fist, etc
2. Weapon should be categorized by weapon speed. Weapon with different base speed are NOT similar.
For example. 1.8: Dagger, 2.6: Fist Weapon, one-hand mace, etc, 3.6: Two-Hand mace, etc.
3. Both item don't have any unique equip restrictions and Both item are not tournament gear (Special item that can be purchased by gold and can only be used in PvP War Games).
4. The item being rolled has no special effect (not set piece as returned by ```GetItemInfo```, no ```ITEM_SPELL_EFFECT, ITEM_SPELL_TRIGGER_ONEQUIP, ITEM_SPELL_TRIGGER_ONPROC, ITEM_SPELL_TRIGGER_ONUSE (Effect: %s, Equip:, Chance On hit:, Use:)``` in the item tooltip)


Do you have any case that my definition will cause false autopass?

Issue #149, opened 2018-01-23
Send if appearance is collected.


SafeteeWoW:
This is probably for v2.9.
1. As a ML, I want to see if a raid member has collected an appearance so I can easily distribute an item that  everyone has passed.
2. The only technical issue is that how I can know if the appearance is collected when the item is a set token. There is no API to relate the tier token to the appearance. Some data is needed in the LUA file, but it is hard to export this data without lots of manual work.
3. For non-token, such info can be easily retrieved by tooltip parsing.

evil-morfar:
I image `C_Transmog` and `C_TransmogCollection` can be helpful. `C_TransmogCollection.IsNewAppearance` sounds promising.

In ExtraUtilities, I always wanted to know how many set pieces a candidate has equipped, but I haven't found a way to do it without hardcoding data. For appearances, the above will probably work though.

I'm reluctant to include this in RCLootCouncil. To me, this sounds like a niche thing, which I've purposely pushed to EU, where it definitely could go. I also once considered doing a "Gear Manager" module, which would track gear the players are wearing, and register how they got the gear (which response, other places, etc) along with other statistics. If that was ever made, then this idea would fit nicely in there as well.

SafeteeWoW:
The problem to include it in EU is that ML would need raid members to also install EU to acquire the appearance information, and I think in the current EU, you are not requiring raiders to install EU?

SafeteeWoW:
I know C_Transmog and C_TransmogCollection, but those API uses appearance set id, which is not the same as the set id returned by GetItemInfo, which is not returned any way if the item is a set token, not an actual gear.

Another problem is that one set token is associate with the appearances of 4 classes, and token with different difficulty offers different appearances. If we do a hard coding, it's a lot of work.

evil-morfar:
There's a few things in EU that doesn't require others to install it, but most of the new columns does. It's relatively simple for a guild to require people to install several addons if they so desire.

After a bit of research, I don't think there's anyway to gather this info for set tokens without hardcoding the data.

SafeteeWoW:
See PR #154

PR #154, opened 2018-02-16
Send Transmog info

SafeteeWoW:
See also issue #149 

1. I don't think this feature should be put in EU, because it happens very frequently that I need to know the transmog info so an non-upgrade piece for the raid will not be wasted.
  + I have a "Transmog" button for the raid. But most people simply clicks pass if they don't need the item, because some people don't actually care about the transmog collection. But as a ML, I don't want an item simply get wasted and I don't want to spend time asking people if anyone possibly doesn't have the appearance.
  + There is no way to know this in-game without asking the raider. There is always some other way to obtain the data of any current feature in EU, including bonus loot (by raid chat, though not 100% reliable), RCScore (by Details/Skada/Recount), etc).
   + > It's relatively simple for a guild to require people to install several addons if they so desire.
   + ^ Although it's true for some people, it's very inconvenient to put some very useful features in a plugin of the addon. And don't expect people to be smart enough to install all required addons. More addons required, less likely. It adds another complexity to download and active the plugin, and lots of people will have no idea of the useful feature in the plugin because it is not in the core of the addon. Use DBM as an example. Currently DBM consists several plugins, including AntorusBurningThrone, Argus, BrokenIsles, Brawlers, etc. Each of these has its own TOC file and can be released as a separate plugin on Curseforge. However, DBM author packs all plugins relevant to the current expansion under the main addon DBM. Imagine all those plugins are released separated, it will be very painful, especially for the new-comers of the game. BigWigs does the same thing. RCLootCouncil is similar to BigWigs and DBM that addon communication is required, therefore addon compatibility between raiders is very important. More steps needed to properly configure the addon, less likely it works. Therefore, I really don't like the existence of Extra Utilities right now.
  + I know combine EU into core takes more memory and CPU, but it can be easily solved by adding settings into RC to enable/disable feature. Unlike EPGP module, EU has nothing major and can be easily combined into the core. (Why DBM still has TrialofValor/EmeraldNightmare modules in the core? Because it is more painful to have no DBM when you do need it and those modules don't take CPU (loaded on demand))
  + Reasons that EPGP module must be standalone and cant be combined into core.
    1. Much large and complex (240KB, vs EU 60KB, code+locale, no lib). For some features of EPGP modules, I need long text to explain how it works.
    2. EPGP modules is never useful to a guild that does not use EPGP as loot rule.
    3. You, evil_morfar, don't write the EPGP module.
  + ^ The above doesn't happen in EU
  + Reasons why TradeSkillMaster releases all its plugin separately. Currently, TradeSkillMaster is designed for advanced players that it requires player to watch lots of video/stream/Wki to make money out of it. This addon currently not designed to be simple to new players. But RC should be simple to new people into the game.

2. I would like to move BonusRoll column from EU into core, because this is very very important for the loot decision, and this is true for any guild, casual or hardcore, unlike Pawn score, or #sockets equipped.
  + I hope the feature will be improve some time to also show if:
    1. The bonus cant be done because there is no more currency. （cross symbol?）
    2. The bonus roll is canceled. The raider clicks No. (stop symbol?)
    3. The bonus roll is pending. The raider is still making his decision. (funnel symbol?)

3. Now it's time to talk about how this PR works.
+ Raider sends if the he has collected the appearance of the item being rolled to the ML.
+ ML shows the info in the Mog column in the voting frame.
   1. If the item has no appearance (Artifact relic, neck, ring, trinket), show nothing.
   2. If the item should be autopassed by the candidate, show nothing, because the appearance cannot be added to the candidate's collection even if he receives the item.
   3. Otherwise, if the candidate has collected its appearance, show green check symbol.
   4. Otherwise, if the candidate has not collected its appearance, show red cross symbol.
   5. Otherwise, if we don't know (because candidate hasn't installed the new version of RC), show question mark symbol.
   6. When ML mouseover thoses symbols, it shows "Apperance collected", "Apperance not collected" or "Apperance unknown"

4. Current problems
+ Nothing send and shown for the tier tokens. I may implement in the future when I find an easy way to generate hardcode appearance data for tier tokens.
+ It could be little bit confusing that "Appearance collected" is displayed as green symbol. Because normally in RCLootCouncil green means you need the item, which is contradict to "Appearance Collected". Therefore, I add the tooltip to clarify this.

### DIFF
--- a/Locale/enUS.lua
+++ b/Locale/enUS.lua
@@ -259,6 +259,7 @@ L["ml_sees_voting_desc"] = "Allows the Master Looter to see who's voting for who
 L["Modules"] = true
 L["module_tVersion_outdated_msg"] = "Newest module %s test version is: %s"
 L["module_version_outdated_msg"] = "The module %s version %s is outdated. Newer version is %s."
+L["Mog"] = true
 L["More Info"] = true
 L["more_info_desc"] = "Select how many of your responses you want to see the latest awarded items for. E.g. selecting 2 will (with default settings) show the latest awarded Mainspec and Offspec items, along with how long ago they were awarded."
 L["Multi Vote"] = true
--- a/Modules/votingFrame.lua
+++ b/Modules/votingFrame.lua
@@ -51,7 +51,8 @@ function RCVotingFrame:OnInitialize()
 		{ name = L["Votes"], 	DoCellUpdate = RCVotingFrame.SetCellVotes,		colName = "votes",	sortnext = 7,		width = 50, align = "CENTER", },				-- 10 Number of votes
 		{ name = L["Vote"],		DoCellUpdate = RCVotingFrame.SetCellVote,			colName = "vote",		sortnext = 10,		width = 60, align = "CENTER", },				-- 11 Vote button
 		{ name = L["Notes"],		DoCellUpdate = RCVotingFrame.SetCellNote,			colName = "note",								width = 50, align = "CENTER", },				-- 12 Note icon
-		{ name = _G.ROLL,			DoCellUpdate = RCVotingFrame.SetCellRoll, 		colName = "roll",		sortnext = 10,		width = 50, align = "CENTER", },				-- 13 Roll
+		{ name = _G.ROLL,			DoCellUpdate = RCVotingFrame.SetCellRoll, 		colName = "roll",		sortnext = 10,		width = 50, align = "CENTER", },
+		{ name = L["Mog"],			DoCellUpdate = RCVotingFrame.SetCellMog, 		colName = "mog",		sortnext = 10,		width = 30, align = "CENTER", },				-- 13 Roll
 	}
 	-- The actual table being worked on, new entries should be added to this table "tinsert(RCVotingFrame.scrollCols, data)"
 	-- If you want to add or remove columns, you should do so on your OnInitialize. See RCVotingFrame:RemoveColumn() for removal.
@@ -351,7 +352,8 @@ function RCVotingFrame:Setup(table)
 				note = nil,
 				roll = nil,
 				voters = {},
-				haveVoted = false, -- Have we voted for this particular candidate in this session?
+				haveVoted = false, -- Have we voted for this particular candidate in this session?,
+				mog = nil,
 			}
 		end
 		-- Init session toggle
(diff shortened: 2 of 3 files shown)


evil-morfar:
1. It's hard to compare RCLootCouncil to other addons like DBM and TSM. The main difference with DBM and RC is DBM has a cleanly defined usage: whatever info you need to know about a boss. RC does alot of different things, and there's no clear definition of when it's "enough". TSM has the same "issue" and thus is split up in several different parts.  
  The main reason for EU's existence is exactly to avoid having too many things thrown at new users. As it is, there's a lot of settings and features a lot of people don't actually use. Furthermore there's no reason to use resources on scarcely used stuff, hence the opt-in nature of EU (which unlike the other addons can't be load on demand).
As for the transmog info - if people can't be bothered clicking a button, do you really think they'd appreciate the implications of adding this? Not to mention it doesn't matter if people don't care (although I still like it).
2. So would I. As you say, unlike the other stuff in EU bonus rolls is really quite essential. The primary reason I didn't add it to the core addon is because I don't want to add any more columns without having a system for disabling/resizing columns (which currently only exists in EU). Again it comes back to the user experience, and adding more stuff with enable/disable options does not improve that (which is also why I haven't prioritised moving that from EU). This is something I'd like to change in a future version though (maybe before 3.0).
I like your ideas, and already envisioned doing that - I just haven't had much time for it.
3. I still like the idea, but I also still don't think it belongs in the core. It would add yet another comm layer, which the current system probably can't really handle many more of - not to mention the root problem is solved by people clicking a button.
4. Luckily tier tokens won't be a problem in 8.0 :)
As for the green "meaning" - I wouldn't dwell to much on that. A lot of people changes the colors (and a equally large part doesn't seem to care at all), and with a check mark that would be an entirely new symbol which holds more meaning than a text color (it's also used in other addons for transmogs, not to mention it indicates "ready" in ready checks).

SafeteeWoW:
1. > As for the transmog info - if people can't be bothered clicking a button, do you really think they'd appreciate the implications of adding this? Not to mention it doesn't matter if people don't care 

The goal of this PR is not doing auto button clicking for raider who care or don't care about transmog. Non-council cant see the voting frame, thus cant really see the changes by this PR. This PR is for the ML who never wants an item to be wasted.

2. RCv3 should have a separate file for scrollcols. There are too many duplicate code to handle scrollcols currently. lib-st is super old. I think we should add some useful custom functions into it.

3. It would add yet another comm layer, which the current system probably can't really handle many more of - not to mention the root problem is solved by people clicking a button.
  + The comm adds 2n+6 bytes, not too much
  + I have considered to make this feature as auto rolling transmog when clicking pass if appearance not collected. But I think ML should differ ppl who really wants the mog and ppl without the mog who don't need the mog, so I give up the idea. But the problem in my group is that most ppl don't care about mog, so most of them click pass instead of mog, so I want to differ among those ppl who don't care about transmog, who has the apperance, and who don't have appearance, so the gear is not wasted. So currently, when everyone passes, I simply throw it to some random guy, and later he says I already have the appearance and I am absolutely don't need it, is there anyone else who may need it or its appearance? I don't want this to happen and this happen fairly frequently that I don't want to spend time asking who may need the appearance.

4. But most ppl, at least 50% I think, uses the default response color. Although I agree I shouldn't worry too much about it. P.S.: Is the default response of relic buttons still grey?

evil-morfar:
1. Except it's very possible for non-council members to see the voting frame, and either way the council members still will. And no, the column shouldn't only be shown to the ML. This PR is for people interested in transmogs - imo an item is not wasted if no one wants it.

2. Yup. It's definitely possible to do drag and drop - it's just very time consuming to do, hence the current implementation. Same goes for click resizing, albeit easier.

3. My point here was just that the original problem is already solved, and again, imo an item is not wasted if no one wants it/can use it. 
The only way I see this being an issue is when a ML really cares about it, but can't get people to install another addon or don't think the issue is big enough to ask them. That's not ideal, but I care more about not bloating the core addon however.

4. Yes.

As for the EU comments:
* While `Legendaries` surely were more useful in the begging, Amanthuls does make it really useful again.
* `ilvlUpgrade` just a remnant.
* `Titanforged` Why? It has always registered correctly every time I've checked. Also note this intentionally included warforged items as well.
* `Sockets` I intended it to only include socketed gems. One could argue the name or intent is wrong, but the idea was to show the amount of sockets benefiting the candidate. True it misses the 3rd potential socket.
* `Traits` my guild requires a specific number of traits on mythic. Not to mention it makes it easy to see who's slacking.
* `Pawn` I expect people using it would set it up correctly, and be aware of the limitations of stat weights. Otherwise the default behavior **is** for people to send their custom stat weights to the ML (or at least the calculated score). When first installing EU, it does not register ones current default stat weights, but that's on the todo (and also the primary reason why EU is still tagged as beta). Personally I don't use stat weights to determine gear overall, but still find it very useful for comparing individual pieces (for which I use pawn all the time). And while you're right stats are worth less the more you have (which is included in the calculation, if people have set it up), it's still very usable for comparisons as you can quickly filter responses based on it.

* `RCScore`. This has yet to be released, as I'm not 100% happy with the current formula. The intent with this is to bring a rating (not unlike warcraftlogs ilvl%, see below) into the game. This could be very useful for comparing player performances, further reducing the guesswork of where a gear would have most effect. Of course this will never be 100% precise due to the use of in game damage meters (although Details! is not far off) and the users setting of those, but the calculated scores are relatively imprecise, thus usable.

The warcraftlogs ilvl% is percentile based, and while it could be duplicated almost 100% in-game, that would require an up to date extract of the data the percentiles are based of (i.e. every single log from the past 2 weeks). It's doable using their API, and it might even be possible to estimate the results to avoid having to update frequently, but all in all not something I'm looking for.
Instead, a much easier approach is to calculate how far a candidate is from the proven max dps, based on ilvl. In general, this would be `dps / maxDpsAt(ilvl)` % which is very easy to create a formula for.

The current formula is an approximation of the average dps/hps made by the top 99 percentile on warcraftlogs. It *does* take ilvl into account, and it's actually a pretty good approximation (only about ~0.5% off). The problem with it is specs are usually +-10% off each other which makes it very imprecise when comparing specs far from each other on the dmg meters. Furthermore there's a lot of discrepancy between fights, which makes an average estimate vary wildly from fight to fight.
To this end I've made a small program that pulls the top n-percentile (most likely 99 due to data) for each fight for every ilvl of every spec. This is linearly approximated (on average this has about 0.15% error rate) and extracted as an invokeable lua function.


**But most importantly:** shared by all of the above: they're completely optional. Don't like, don't use it. The only reason they're in EU is because I believe not everyone will use them.

Issue #155, opened 2018-03-06
Autopass when saved or didn't tag the boss


SafeteeWoW:
By comparing the raid lockout before the encounter and after the encounter, whether or not the player is eligible for the loot can be known.

```GetMasterLootCandidate``` is another way to check this, but this occasionally changes when a raider leaves instance and is still eligible for the loot, at least when I tested it in Patch 7.3 (before 7.3.5) I have never done any loot tests since 7.3.5, because I can no longer master loot in old dungeons.

1. If the player is not locked to the boss and is locked to the boss after it is dead, the player is eligible for the loot, respond to ML normally.
2. If the player was locked to the boss before the instance, autopass by sending "SAVED" to the ML.
3. If the player wasn't locked after the encounter, autopass by sending "OUT" to the ML.
4. We only do the above check when the item is BoP.

Such information can be gathered from the following API:
```GetSavedInstanceInfo, GetNumSavedInstances, GetSavedInstanceChatLink```
```RequestRaidInfo RequestLFDPlayerLockInfo```
Whether or not a boss is saved can be known by parsing saved instance chat link. (https://wow.gamepedia.com/SavedInstanceChatLink)
For more reference, read FrameXML/RaidFrame.lua and the addon SavedInstances

There are several challenges for this feature:
1. (a)Award Later list should maintain the boss which the item belongs to, otherwise this does not work if ML awards many items later, or (b) we hardcode the boss-item relationship. This is probably a better bug-free version. I think addon has implemented this feature, and it's not hard for me to dump this information from the Encounter Journal, just need some disk space to store such information in the source code.
2. The recent saved instance information should be recorded in the SV for every raider, so the autopass by lockout can be sent correctly if ML does award later.
3. There could be a delay between the boss kill and the registration of  lockout. This should be tested by experiment. If the player wasn't locked before the encounter, and still wasn't after the encounter, but see the ENCOUNTER_END event with a kill, we can assume the eligibility for the loot????
4. Privacy issue. DBM can fetch all lockout information from raiders too, but raider won't send this information unless confirm by clicking the dialog. Therefore, RCLootCouncil never sends all lockout infomation. It only sends information for the most recent encounter. Since most of the time, ML can know the lockout information by reading the master loot list anyway, it is not a privacy leaking.

Why this feature:
[...]

PR #158, opened 2018-03-24
Show out of instance in voting frame

SafeteeWoW:
Only done LUA syntax check. Haven't run in-game. Missing localization entry.
I have talked about this in #155 

I am considering to put "Out of instance" in the response column instead of name column.

### DIFF
--- a/Modules/votingFrame.lua
+++ b/Modules/votingFrame.lua
@@ -968,9 +968,9 @@ function RCVotingFrame.SetCellResponse(rowFrame, frame, data, cols, row, realrow
 		or ((not IsInInstance()) and UnitPosition(Ambiguate(name, "short")) ~= nil) then
 		-- If the current player is not in an instance, mark as out of instance if 1st return of UnitPosition is not nil
 		-- This function returns nil if the raider is in any instance.
-		text = text.." (Out of instance)"
+		response = response.." (Out of instance)"
 	end
-	frame.text:SetText()
+	frame.text:SetText(response)
 
 	frame.text:SetTextColor(addon:GetResponseColor(lootTable[session].candidates[name].response, isTier, isRelic))
 end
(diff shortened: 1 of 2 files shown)


Modules/votingFrame.lua
```diff
@@ -930,6 +935,13 @@ function RCVotingFrame.SetCellName(rowFrame, frame, data, cols, row, realrow, co
 	local name = data[realrow].name
 	frame.text:SetText(addon.Ambiguate(name))
 	local c = addon:GetClassColor(lootTable[session].candidates[name].class)
+	if (IsInInstance() and select(4, UnitPosition("player")) ~= select(4, UnitPosition(Ambiguate(name, "short"))))
+			-- Mark as out of instance if the current player is in an instance and the raider is in other instancemap
+			or ((not IsInInstance()) and UnitPosition(Ambiguate(name, "short")) ~= nil) then
+			-- If the current player is not in an instance, mark as out of instance if 1st return of UnitPosition is not nil
+			-- This function returns nil if the raider is in any instance.
+			c = c.." (Out of instance)"
```
evil-morfar: `c` is the color table, not the text. 
I'm still not sure it should be stored in the name, see [#158](https://github.com/evil-morfar/RCLootCouncil2/issues/155#issuecomment-375966623).

SafeteeWoW:
1. I have moved the text to the end of response
2. ```GROUP_ROSTER_UPDATE``` does not work. Instead. ```UNIT_PHASE``` fires whenever any other group members enter/exit an instance. ```ZONE_CHANGED_NEW_AREA``` fires when the current player enter/exit an instance. 
3. Tested in game.
4. The next step is to test whether ```UnitPosition``` returns the correct result immediately after ```UNIT_PHASE``` or ```ZONE_CHANGED_NEW_AREA```

SafeteeWoW:
This feature could be in v2.10

Issue #160, opened 2018-03-26
Uncached lootTable items


evil-morfar:
There's a small potential for throwing away data with the current lootTable caching system. Consider this log from a recent raid of mine:

```lua
"20:15:50 - Comm received:^1^SlootTable^T^N1^T^N1^T^SequipLoc^SINVTYPE_HAND^Sawarded^b^Slink^S|cffa335ee|Hitem:152012::::::::110:104::5:3:3611:1512:3337:::|h[Molten~`Bite~`Handguards]|h|r^Srelic^b^Stexture^N1605809^SsubType^SPlate^SlootSlot^N3^Sclasses^N4294967295^Sname^SMolten~`Bite~`Handguards^Sboe^b^Silvl^N970^Squality^N4^t^N2^T^SequipLoc^SINVTYPE_TRINKET^Sawarded^b^Slink^S|cffa335ee|Hitem:151974::::::::110:104::5:4:3611:3618:1497:3336:::|h[Eye~`of~`Shatug]|h|r^Srelic^b^Stexture^N463857^SsubType^SMiscellaneous^SlootSlot^N2^Sclasses^N4294967295^Sname^SEye~`of~`Shatug^Sboe^b^Silvl^N955^Squality^N4^t^N3^T^SequipLoc^S^Sawarded^b^Slink^S|cffa335ee|Hitem:152291::::::::110:104::5:3:3611:1492:3336:::|h[Fraternal~`Fervor]|h|r^Srelic^SLife^Stexture^N459025^SsubType^SArtifact~`Relic^SlootSlot^N1^Sclasses^N4294967295^Sname^SFraternal~`Fervor^Sboe^b^Silvl^N950^Squality^N4^t^t^t^^ (from:) (Barrow) (distri:) (RAID)", -- [1356]
"20:15:50 - Comm received:^1^SlootAck^T^N1^SEliarra-Ravencrest^N2^N258^N3^N961^N4^T^Sresponse^T^N1^B^N2^B^t^Sdiff^T^N1^N25^N2^N5^N3^N0^t^Sgear1^T^N1^Sitem:152155::::::::110:258::5:3:3611:1487:3528^N2^Sitem:151962::::::::110:258::3:3:3610:1492:3337^t^Sgear2^T^N2^Sitem:151955::::::::110:258::5:3:3611:1492:3336^t^t^t^^ (from:) (Eliarra) (distri:) (RAID)", -- [1357]
[...]

PR #164, opened 2018-03-31
RCItemUtils Library

SafeteeWoW:
I create a RCItemUtil library to manage the code to fetch item information
The purpose of this library.

+ Significantly reduce the redundant code for item information handling.
+ Significantly reduce the number of calls to WoW Item APIs and tooltip display.
+ Strong error checking makes it much harder to make coding mistakes
+ No longer need to the memorize the order of returned values of ```GetItemInfo```
+ Less headache for handling uncached items by setting default values for uncaching items.
+ Storing item information in the loot table is no longer needed.

It is still incomplete.
But can you read this file and give me some suggestions? Sorry for lack of code comments.

The main idea is that you can use ```GetItemAttr``` to get one attribute of the item, for example ```GetItemAttr(ItemString, "name")```. ```CacheItem(item)``` will cache all item attributes in a table, so we no longer needs to call any APIs in ```GetItemAttr```. I have also defined some functions to conveniently define item attributes.

---
New discovery:
The api ```DoesItemContainSpec(itemLink, classID, [specID])``` and ```GetItemSpecInfo``` gives item spec information. This api eliminates the need to export trinket data and all other hardcode stuffs for autopassing _*This API also works for artifact relics and armor tokens*_. The former is a bit buggy if the item is not in the encounter journal(but it does work for armor tokens), for example, artifact weapon, but I think it should not matter. The latter only returns spec in the player's class. Not sure why don't I know this before. Are these added recently?


SafeteeWoW:
Btw, this does not have memory problems. Caching all items in my full bags and bank takes 3MegaBytes, about 300 items.

SafeteeWoW:
Due to the complete code restructure, this will be in v2.9, and I will need to write a separate test lua file to make sure nothing is wrong.

SafeteeWoW:
This PR will deprecate all of following entries in the loot table.
name, quality, ilvl, equipLoc, subType, texture, boe, relic, token, typeID, subTypeID, classes.
and client side no longer needs to cache items to fix #160 
WIP
TODO: Delay autopass until items are cached.

SafeteeWoW:
Can you help to make sure how ```DoesItemContainSpec(itemLink/ID, classID, [specID])``` works?
Return value:
true if the item is in the lootspec of the given class/spec. false otherwise
Arguments:
1. itemLink/ID: Required. Accept item link/string/id
2. classID: Required. Class id number ranged from 1-12 (1 is warrior and 12 is demon hunter)
3. specID: Optional. If nil, API returns true if any spec of the class can loot the item. Otherwise, only check the given specID. specID is a three-digit number.

It seems this function does not have caching requirement. Can you help me to check this?
Script I used
```
local t = {}
for id in pairs(_G.RCTokenTable) do
    t[id] = DoesItemContainSpec(id, 11)
    GetItemInfo(id)
end

C_Timer.After(20, function() print("time is up");
    for id in pairs(_G.RCTokenTable) do
        local c = DoesItemContainSpec(id, 11)
        if  not (c == t[id]) then
            print(id, t[id], c)
        end
    end end
)
````
The only print is "time is up". So I think this API does not have caching requirement.

evil-morfar:
`DoesItemContainSpec` [seems to have been in the game at least since 7.0.3](https://pastebin.com/2m0DhzfG).

I once again find myself with very limited time for WoW, so I probably won't be of much help until next week, but it seems like you figured it out?

evil-morfar:
Also, when you say "3 MB" is that the total memory usage of RCLootCouncil, or just the items? If the latter, 300 items at 3 MB that's 10 kB each, which is a lot.

PR #179, opened 2019-12-26
Add search box

mtoebes:
Problem: the history frame only allows you to search one player name at a time. In addition, that player name must be selected from a long scrollable list.

solution: add a search box. 

Player names (excluding the realm name) can be added/removed by clicking on their name from the scrollable list or by typing it into the box. The logic uses string begins with to find matches. This also allows for multiple player exports. 

Bonus: add an option to switch search to be by class or item name (not export enabled)

### DIFF
--- a/Modules/History/lootHistory.lua
+++ b/Modules/History/lootHistory.lua
@@ -16,11 +16,13 @@ data[date][playerName] = {
 	}
 }
 ]]
-local selectedDate, selectedName, filterMenu, moreInfo, moreInfoData
+local selectedDate, filterMenu, moreInfo, moreInfoData
 local rightClickMenu;
 local ROW_HEIGHT = 20;
 local NUM_ROWS = 15;
 local epochDates = {} -- [DateTime] = epoch
+local searchList = {}
+local searchType = "name";
 
 --globals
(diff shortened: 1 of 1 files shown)


evil-morfar:
Interesting. I'm currently on vacation so a full review will have to wait, but I imagine this could go into v2.17 I plan to release for patch 8.3.

evil-morfar:
@mtoebes Did you intend to finish this?

mtoebes:
If you are interested in the idea, I'll resolve the conflicts. Else, I'll close the PR.

evil-morfar:
I am interested

mtoebes:
I merged in the lastest master. should to rdy for another looking at.

Unrelated. I think a guildie ran into some tiny bugs (that I have not seen for myself). 

Can you check that time_remaning isn't meant to be time_remaining in ml_core (causing a arithmetic on nil value error at like 432). 

Also, are the params going into self:ScheduleTimer("LootOpened", count / 10, autoloot, count) in Module::91 correct? I think the autoloot flag is being used for count in the method. This is causing an arithmetic error on local count (a boolean value) at line 87.

PR #180, opened 2020-01-07
JSON export

sebastianbassen:
This PR implements a json export. It will use the following format:

`
[ 
{ 
"player":"Babababa-Skullflame",
"date":"29/12/19",
"time":"20:42:48",
"itemID":16914,
"itemString":"item:16914::::::::60",
"response":"Response",
"votes":0,
"class":"MAGE",
"instance":"Onyxia's Lair-40 Player",
"boss":"Onyxia",
"gear1":"",
"gear2":"",
"responseID":"NOTHING",
"isAwardReason":"false",
"rollType":"normal",
"subType":"Cloth",
"equipLoc":"Head",
"note":"",
"owner":"Onyxia"
},
...
]
`

Never used lua before so if you have any suggestions for improvements please tell me :) I only tested this on WoW Classic, so I would appreciate it if someone can test this on retail.


evil-morfar:
Sorry I haven't got back to you earlier.

Looks promising - have you tested in in-game?
Assuming there's no bugs I overlooked this will go into the next release.

As a side note, creating new strings in lua is "expensive". Lua strings are unique and stored in a global table, which is great when you want to use a single string multiple times, but not so great when creating many of them (concatenations creates new string every time!), and the string itself is massive (i.e all the data). Using `table.concat` gets around that by essentially only creating a single string with all the data in the table. For that reason, it's better to add even a single character to the table than to concatenate it into the final string.

I'm not sure if it *really* matters in this context, but now you're aware.

sebastianbassen:
Hey, thanks for the feedback. I only tested this on WoW Classic. Currently im getting a error message " AceLocale-3.0: RCLootCouncil: Missing entry for 'Standard JSON Output.'" and I don't know why.

Performance seems good, for a history export with over 5k lines of JSON the export string pops up instantly.

evil-morfar:
Are you on a non english client?

evil-morfar:
I fixed the issue as it was so minor. Everything seems to work, and it produced a valid json output of my entire history of ~3000 items.

Thanks for your contribution!

PR #183, opened 2020-02-29
Clearer item loot dialog choices (yes/no -> keep/trade)

Hoishin:
After looting bosses, our guild have a repeated problem of accidentally selecting Yes when they don't want keep the item, or vice versa. They have to tell the raid leader that they didn't mean it after that and manually count it for rolls.

The problem here is that the choice isn't really matter of saying yes and no, but rather they choose if they want to keep it or trade it.

This PR makes the dialog clearer by changing the choices "Keep" and "Trade", minimizing the possibility of such errors.


evil-morfar:
Just FYI, this doesn't work as it's missing the translation phrase you've altered.

Hoishin:
Should be fixed

PR #204, opened 2020-12-29
Dynamically update EJLastestInstanceID

jjholleman:
Added a function to get the latest raid instance id.
EJLastestInstanceID is no longer hardcoded and instead uses the new fuction.

this change makes sure that `/rc fulltest` will always pull items from the latest raid.


evil-morfar:
Interesting choice for a PR, but thanks anyway. This is not exactly urgent with the issues present at the moment, but I'll include it in the next non-bugfix update.

PR #206, opened 2021-01-14
Frame level

enajork:
Modules/Frames should now properly obscure each other (especially their contents) depending on which one has focus.

I did not test this thoroughly but I believe this is the correct solution to the problem.


enajork:
It would also be great to get this fix into the Classic version as well. I see that the Classic version uses 7968cf257ef6cf4619367a575e5251649dfcda63 and might not benefit from these changes immediately. Thank you.

enajork:
Whoops closed by accident. Sorry for spam.

enajork:
I tested the session, voting, and history frames on Retail and Classic. Everything looks good. Some other modules/frames may still need to be tested.

Modules/sessionFrame.lua
```diff
@@ -252,7 +252,7 @@ function RCSessionFrame:GetFrame()
 	f.lootStatus.text:SetJustifyH("LEFT")
 
 	local st = ST:CreateST(self.scrollCols, 5, ROW_HEIGHT, nil, f.content)
-	st.frame:SetPoint("TOPLEFT",f,"TOPLEFT",10,-20)
+	st.frame:SetPoint("TOPLEFT",f,"TOPLEFT",10,-ROW_HEIGHT-10)
```
enajork, line 255: Could just make this -50. No particular reason I did it this way.

evil-morfar:
I've been wanting to look into this for quite a while, so thanks a lot!

Classic is still based on the v2 design, which doesn't have the frames in the same place. Shouldn't be too hard to implement there, as all I more or less copy/pasted the code straight into `Widgets/Frame.lua`.

enajork:
I implemented the fix in v2 first then in v3. Just didn't want to bother creating a separate pull request for v2 (seemed more complicated). Figured you would copy pasta for the Classic version 😄 

Thank you so much!

PR #209, opened 2021-05-21
Updates for TBC Classic

corylation:
With TBC classic the WoW client now requires the BackdropMixin to be added to a CreateFrame call if the frame's parent doesn't include a Backdrop. All updates herein are designed to solve that API change.

I tried to make minimal commits to update the addon to support TBC Classic by doing this:

- Start from the current commit the submodule is set to in the RCLootCouncil_Classic Repo
- Update bundled libraries
- Update RCLootCouncil code that requires the BackdropMixin added to CreateFrame

The vast majority of the updated code here are library updates. The changes to RCLootCouncil's code are minimal.


evil-morfar:
Thanks. I originally intended to update the Classic version to be up-to-date with the current v3 used for retail, but as I have limited time I'm going with your approach for now.

PR #212, opened 2022-02-10
Add option to scale window

danielandersson:
Simply added an option to scale the size of the windows.

Feel free to change target from `master`


evil-morfar:
You can already do it with ctrl+scrollwheel on any window. I see no reason to have a separate option for it.

On a side note, the only reason windows are big by default is because I originally scaled them on a less than 1080p monitor while failing to realize I had my UI scale below 1. One day I might get around to reset the scale.

PR #214, opened 2022-03-24
Fix hyperlink formatting for TSV export

Yttrium-tYcLief:
The `=HYPERLINK` function in Excel uses comma delimiters, not semicolon delimiters. This commit means that when you paste directly into Excel, the desired cells are correctly formatted, when previously the exported data just displayed the raw text of the broken function.

See more: https://support.microsoft.com/en-us/office/hyperlink-function-333c7ce6-c5ae-4164-9c47-7de9b76f577f


evil-morfar:
Not sure why it was ","'s, maybe some localization stuff. Anyway, thanks for the PR!

Issue #236, opened 2023-05-19
Autopass | Don't pass on transmoggable items


Urtgard:
I would like to have an extra option to not auto pass on items I could use for transmog. E.g. a warrior can learn bow appearances.
Actually this should be two options to distinct between source or appearance.

I could add this feature myself if you'll accept a pr for this.

evil-morfar:
If you can create it, I'd be more than happy to include it. I'm not entirely sure how to determine whether a player can use a certain item for transmog or not

PR #237, opened 2023-05-26
Autopass | Don't pass on transmoggable items

Urtgard:
Implementation for #236 

By default auto pass now does not pass on unkown transmog appearances.
Optional players can choose to not pass on unkown transmog sources.

I did not add any tests because I could not get the busted setup right in a short time.


evil-morfar:
Awesome! Tests aren't really setup for this scenario, and everything would need to be mocked anyway, so the tests would only really be testing logical/syntax - don't bother with it.

Two comments though:
1. Would anyone really need to have an item that they already has the transmog for? Seems to me it would be simpler to only keep the `autoPassTransmogSource`.
2. I'd prefer if auto passing transmog is enabled by default, so that the common raider doesn't experience any changes simply by updating.

evil-morfar:
3. Since `autoPassTransmogSource` relies on `autoPassTransmog` being `false`, I think it'd be better to disable the option when the latter is `true`. I.e. add `disabled = function() return self.db.profile.autoPassTransmog end`

Urtgard:
Made the two changes based on your comments.

evil-morfar:
Thank you for the contribution - it'll be included in the next update (which will be soonish (tm)).

Issue #187, opened 2020-05-14
Time zone differences not accounted for


mcd8604:
I noticed tonight that the history timestamps are saved based on the master looter's timezone. We ran two raids tonight and changed master looter. The second ML happened to have an earlier timezone and the gear drops got interspersed with the prior raid:

![rclootcouncil_timestamp_issue](https://user-images.githubusercontent.com/2568468/81893568-f1e42c00-957b-11ea-9900-4982978305fb.png)
The two highlighted items were from BWL and the items in between came from MC which we ran an hour later.

Is it possible to save the timezone along with the timestamp?

evil-morfar:
I'm working on revamping the time system. I'm treading carefully as it could have some backwards compatibility issues, but I'll most likely change all awards to be based on server time.

Orphis:
When you do that, make sure to use an ISO format for dates, dd/mm/yy is totally ambiguous when you try to import it in any software and doesn't sort nicely. YYYY-MM-DD works better.
Also, there's no reason to not have the time in the same field for even easier sorting: YYYY-MM-DDThh:mm:ss

johnbaker:
I've run into this in an external app I am working on that is based on RCLootCouncil exports.  The exported timestamps would be more useful they were based on server time or if the timezone information was included with the export.

johnbaker:
@evil-morfar is this still on the roadmap?  Could there be a shorter-term step where the `serverTime` is added to SavedVariables and handled via import/export so we consume the data?  If this is acceptable, I'd even be willing to throw together a PR.

evil-morfar:
> @evil-morfar is this still on the roadmap? Could there be a shorter-term step where the `serverTime` is added to SavedVariables and handled via import/export so we consume the data? If this is acceptable, I'd even be willing to throw together a PR.

It will be the next "big" thing I'll be working on, but I can't give an eta. I'd be happy to include a server time field (maybe except for in the csv export, as that requires the import to be updated as well). Even better if you'd create it :)

evil-morfar:
Implemented in v3.15.0

PR #262, opened 2025-01-15
adds serverTime and field on json export

johnbaker:
The following issue details the current issue with the date and time field in the data:
https://github.com/evil-morfar/RCLootCouncil2/issues/187

This pull request doesn't solve the display issue as care would need to be taken with old data (without the actual serverTime value) and could lead to confusion.  However, as a short-term solution for applications that may either parse the RCLootCouncil.lua saved variable or use the Json export, this pull request adds a "serverTime" field to the loot data using the GetServerTime() function which is synchronized across clients and not effected by the current user's system clock or time zone.

I didn't include the EQXML export as the <enter> and <leave> timestamps are only based on the day (and not time) and dynamically swapping between the date/time epoch and the actual server epoch based on its existence may not desired.  I can add this if requested.


evil-morfar:
Thanks for your PR, but I'm not going to merge it currently, as a rework of everything time related is on my schedule for patch 11.1. This includes switching time stamps to be server time based, and compared to this, doesn't need extra data storage.

I'm gonna leave this as is in case anyone needs it, and with a promise of implementing it for patch 11.1 if for whatever reason I'm not able to do my planned implementation.

evil-morfar:
Added in v3.15.0

PR #264, opened 2025-03-07
Enhanced MLDB.lua stability and error handling

joshjau:
Fixed some weird errors in MLDB.lua that would pop up when certain add-ons using Ace3 were loaded. While tracking down the issue, I found and fixed problems with type checking in ipairs() calls and variable shadowing that were causing conflicts. I also added proper error handling, created helper functions to make the code easier to read, and optimized how data is handled throughout the file. These changes should make everything more stable without changing how the addon actually works I believe, but I could be wrong.

Tested in game, worked fine. I added additional code comments as well for clarity.


evil-morfar:
How exactly are these "errors" occurring? I've never seen anything close to errors related to this, and even if there where this is just a bandaid fix for something more serious occurring elsewhere.

I'm really just seeing a lot of expanding code, where the majority *does not* improve readability imo.

Classes/Data/MLDB.lua
```diff
@@ -79,11 +104,26 @@ function MLDB:IsKey(val)
 end
 
 function private:ReplaceMLDB(mldb, replacement_table)
+    if type(mldb) ~= "table" or type(replacement_table) ~= "table" then
+        return {}
+    end
+    
     local ret = {}
     for k, v in pairs(mldb) do
```
evil-morfar, line 112: Everything above and in this function could be replaced by updating `pairs(mldb)` to `pairs(mldb or {})`. But again, there would be an issue elsewhere causing this

Classes/Data/MLDB.lua
```diff
@@ -93,60 +133,100 @@ function private:ReplaceMLDB(mldb, replacement_table)
...
-            if i > db.buttons[type].numButtons then
-                break
-            end
-            if
-                not addon.defaults.profile.responses[type] or
-                    db.responses[type][i].text ~= addon.defaults.profile.responses[type][i].text or
-                    unpack(db.responses[type][i].color) ~= unpack(addon.defaults.profile.responses[type][i].color)
-             then
-                if not changedResponses[type] then
-                    changedResponses[type] = {}
```
evil-morfar, line 111: This is the only place where I'd argue readability is somewhat improved, but I'm not sure 40+ lines extra is an improvement

evil-morfar:
But on the bright side, you did reveal an error in my implementation:
image

PR #273, opened 2026-03-21
fix: Updated PlayerKnowsTransmogInternal to properly get data from C_TransmogCollection.GetAppearanceSourceInfo

peti446:
Hi,

So in Mdnight the function [C_TransmogCollection.GetAppearanceSourceInfo](https://warcraft.wiki.gg/wiki/API_C_TransmogCollection.GetAppearanceSourceInfo) was changed to return a table meaning that the old way using `select(5,...);` is incorrect and we need to change it to access .isCollected without the change we will never auto pass on items we already have the transmog.

Note: To test this you need uncheck "Auto pass transmog" from the config and not to have the CanIMogIT addon as it will otherwise go trough that path.


evil-morfar:
Cool - thanks :)

PR #277, opened 2026-07-12
Feat/session responses

groeterud:
New version of https://github.com/evil-morfar/RCLootCouncil_Classic/pull/65 into the core addon. 
Ties data sync to ML award and simply expands history data with responses. 

/rc history now looks like this, with a dice icon next to awards with saved responses  
image  

Click to reveal responses with notes etc:   
image


evil-morfar:
Good job - looks fine.
With some luck (time permitting) this can go into the 12.1 update when that launches in a few weeks, then into Classic. If there's no breaking changes (will need to investigate the exact impact of the updates comms) it could go into classic earlier.

Speaking of comms, I'll need to do some extensive testing to see exactly what the impact is, not to mention how much storage this takes up and how it affects exports. I might just require this to be opt-in on the reciepient side (maybe you can add this?) to avoid overly large SV's for the average raider that doesn't care about this (bigger files = larger load times).

groeterud:
@evil-morfar added setting now, defaulted it to on - change as you feel :)

evil-morfar:
Byte size for a history comm after compression:
image
Will definitely need some optimizing for it to be on by default, but I reckon I can reduce it at least 20%

evil-morfar:
I'm running out of ideas, but I managed to get a history comm down to just 1 extra message instead of the original 2.
These tests on all 5 increment raid sizes, repeated 100 times and averaged out. There's a 40% chance for a autopass/pass response.
image

Best case specifics:
image
Test file is [here](https://gist.github.com/evil-morfar/86425134f0589447b7312d19ece14ff5) if you want to play around with it.

evil-morfar:
@groeterud I've implemented the comms optimizations, making the backend part more or less feature complete.
Are you able to test it? I know you play classic, but replacing the RCLootCouncil files in the classic version will make it work. Otherwise I can make a build for you.
Note that the changes I made are not backwards compatible

groeterud:
> @groeterud I've implemented the comms optimizations, making the backend part more or less feature complete. Are you able to test it? I know you play classic, but replacing the RCLootCouncil files in the classic version will make it work. Otherwise I can make a build for you. Note that the changes I made are not backwards compatible

I've been running my side of these commits locally since the 13th. I'll pull down the changes and run it locally. I only manually patched in the comms stuff, not the 12.1 API stuff - since I'm not certain if that is live in MOP

groeterud:
Identified a bug in non-winner responses that contained an '@', added a fix for it. 

Tested OK locally with /rc test. Got raid where I'm ML later on tonight, will report if anything breaks

groeterud:
Testing today, everything fine - had one bug when someone submitted an empty note containing a blank space: " "

Adressed in last commit

groeterud:
marked as ready @evil-morfar

evil-morfar:
Awesome. The Ace3 changes should be backwards compatible.
I still have a few things to add before this is ready to go

evil-morfar:
@groeterud Overhauled the design to make it easier to use. Not to sure about the hover thing, but please let me know what you think.
image

 Only two things remains to be decided:
- Default on or off. I'm leaning to off.
- Separate option for sending? Not sure..

evil-morfar:
In the end I decided to make a new option for sending the data (default off) and left as on - this way only the ML needs to make a choice, while all others can benefit from it, and people concerned with increased storage can opt out.

